PR #17, opened 2015-04-27
Add Keywords to desktop file as recommended

innir:
It's recommended to have Keywords in the desktop file as Unity, Gnome Shell etc. rely on them for proposing applications to users, see: https://wiki.gnome.org/Initiatives/GnomeGoals/DesktopFileKeywords


manisandro:
Thanks!

PR #18, opened 2015-04-27
Don't use Qt macros in GTK+ source

innir:


manisandro:
Oh whoops, good catch! Thanks!

PR #19, opened 2015-04-27
prctl is Linux specific

innir:
The function prctl() is Linux specific, it's not available on all Unix Systems, especially not on *BSD. That makes gImageReader fail to build on kfreebsd. See https://buildd.debian.org/status/fetch.php?pkg=gimagereader&arch=kfreebsd-amd64&ver=3.0.1-1&stamp=1430084023 for reference.


manisandro:
Thanks again!

PR #51, opened 2016-02-19
Sort lists to make build reproducible

innir:
The lists produced by FILE(GLOB ...) depend on the ordering of readdir(). To avoid this the lists are sorted to make build reproducible.


manisandro:
Thanks!

PR #57, opened 2016-03-03
On ARMEL std::exception_ptr is missing - QT part of the fix.

innir:
ARMEL's CPU default is too old to support std::exception_ptr. As Debian sill builds for ARMEL it fails to build without.


manisandro:
Somehow missed this PR. Thanks!

PR #69, opened 2016-06-21
Correct and extend Ukrainian translation

yurkobb:
Fix a couple of existing strings and translate most of the untranslated ones.


manisandro:
Thanks!

PR #104, opened 2016-09-30
Update Chinese translations

timothytylee:
In addition to updating existing Traditional Chinese translations, Simplified Chinese translation has been added for completeness.


manisandro:
Thank you!

PR #112, opened 2016-10-16
[GTK] OutputEditorHOCR: Fix removeCurrentItem

danpla:
We shouldn't clear the m_currentItem, because m_itemStore->erase triggers
selection change.

Fixes crash caused by an invalid iterator.

PS: Is there any code style guidelines for contributors for cases like `if(...) {`?


manisandro:
Thanks!

Can you add a quick comment along the lines of `// m_currentItem updated by m_itemView->get_selection()->signal_changed()`?

Code style: not really, and I'm not always 100% consistent myself... Ideally I should define an `astyle` rule and provide an auto-reformatting script.

danpla:
Done.

manisandro:
Thanks!

PR #113, opened 2016-10-16
[Qt] OutputEditorHOCR: Don't work with null item

danpla:
Fixes crash after clicking on empty area below items.


manisandro:
Thanks!

PR #115, opened 2016-10-17
CMake: Try find_package first to find jpeg

danpla:


manisandro:
For curiosity, is there actually a distro on which libjpeg/libjpeg-turbo ships a cmake file? I've only seen the pkgconfig file or nothing at all up to now.

danpla:
That's one of the standard cmake modules: https://cmake.org/cmake/help/v3.0/manual/cmake-modules.7.html

manisandro:
Ah ok! Thanks.

PR #116, opened 2016-10-18
[Gtk] Displayer::setSources: Don't work with null m_currentSource

danpla:
Fixes crash after drag-and-dropping non-image files.


manisandro:
Thanks!

PR #121, opened 2016-10-21
[Gtk] OutputEditorHOCR: Fix word merging

danpla:


manisandro:
Thanks! - No idea when this actually broke, I'm sure multiselection worked when I implemented this...

PR #125, opened 2016-10-26
[Gtk] SubstitutionsManager::openList: Keep empty lines in Utils::string_split

danpla:
Fixes false error message after loading a list that contains empty pairs.

It probably makes sense to set `keedEmpty` to `true` by default to match Qt's behavior.


manisandro:
Indeed, done in 7b9a864.

Thanks so much for all your bugfixes and improvements!

PR #126, opened 2016-10-28
[Qt] Displayer::getSceneBoundingRect: Fix size deviations

danpla:
We cannot use m_imageItem->sceneBoundingRect() since its pixmap
can currently be downscaled and therefore have slightly different
proportions.


manisandro:
Thanks!

PR #127, opened 2016-10-29
[Qt] Displayer::getImage: Use smooth pixmap transformation

danpla:
Significantly improves quality of rotated images:
pic


manisandro:
Awesome!

PR #178, opened 2017-06-12
Recognize guillemets as quote characters (fixes #177)

### LINKED ISSUE #177: "Keep line break" does not work for all quote characters
Hi,

when using the option "Keep line break if preceded or succeeded by quote" only ' and " are taken into account but not » and «.
I think it's just adding them here: https://github.com/manisandro/gImageReader/blob/master/gtk/src/OutputEditorText.cc#L177-#L180 and there: https://github.com/manisandro/gImageReader/blob/master/qt/src/OutputEditorText.cc#L138-#L141.

Should I just file a PR?

Best,

Philip

innir:
Hi,

here is the pull request. 
I added »« on both "sides" as both can mark opening and closing of quotes (see https://en.wikipedia.org/wiki/Guillemet).

Best,
Philip

### DIFF
--- a/gtk/src/OutputEditorText.cc
+++ b/gtk/src/OutputEditorText.cc
@@ -175,8 +175,8 @@ void OutputEditorText::filterBuffer() {
 			preChars += "\\.\\?!"; // Keep if preceded by end mark (.?!)
 		}
 		if(m_filterKeepIfQuote->get_active()) {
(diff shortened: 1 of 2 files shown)


manisandro:
Perhaps safer to use `\u00AB` and `\u00BB`?

innir:
True, done.

manisandro:
Thanks!

PR #191, opened 2017-10-05
Add djvu reading support

ZaMaZaN4iK:
Hello,

I implemented djvu reading support. I had copied C++ wrapper for djvulibre from Okular project.

Please review my changes and if it's good, merge it.


manisandro:
Hi, looks like you based your work on an old version of the code, can you git rebase against the latest master? Thanks!

ZaMaZaN4iK:
Merged from master. Pls review it.

CMakeLists.txt
```diff
@@ -228,6 +227,7 @@ TARGET_LINK_LIBRARIES(gimagereader
     ${TESSERACT_LDFLAGS}
     ${gimagereader_LIBS}
     ${SANE_LDFLAGS}
+    -ldjvulibre
```
manisandro: djvulibre has a pkgconfig file, it would be better to get include dirs and ldflags from pkgconfig instead of hard-coding them.

gtk/src/DisplayerToolSelect.cc
```diff
@@ -198,7 +198,8 @@ void DisplayerToolSelect::autodetectLayout(bool noDeskew) {
 	// If a somewhat large deskew angle is detected, automatically rotate image and redetect layout,
 	// unless we already attempted to rotate (to prevent endless loops)
 	avgDeskew = Utils::round(((avgDeskew/nDeskew)/M_PI * 180.) * 10.) / 10.;
-	if(std::abs(avgDeskew) > .1 && !noDeskew) {
+    //TODO: CppCheck warning for 'if' below. Is it correct: std::abs(avgDeskew > .1)? bool inside std::abs
+	if(std::abs(avgDeskew > .1) && !noDeskew) {
```
manisandro: You are reverting a previous fix here.

qt/src/DisplayRenderer.cc
```diff
@@ -106,3 +110,32 @@ QImage PDFRenderer::render(int page, double resolution) const {
 int PDFRenderer::getNPages() const {
 	return m_document ? m_document->numPages() : 1;
 }
+
+#include <QDebug>
```
manisandro: This looks like a debug leftover which should be removed.

qt/src/DisplayRenderer.hh
```diff
@@ -23,6 +23,8 @@
 #include <QString>
 #include <QMutex>
 
+#include "Kdjvu.hh"
```
manisandro: You can forward declare the class instead and include the header only in the source file.

qt/src/Displayer.cc
```diff
@@ -107,7 +108,11 @@ bool Displayer::setCurrentPage(int page) {
 		if(source->path.endsWith(".pdf", Qt::CaseInsensitive)) {
 			m_renderer = new PDFRenderer(source->path);
 			if(source->resolution == -1) source->resolution = 300;
-		} else {
+        } else if(source->path.endsWith(".djvu", Qt::CaseInsensitive)) {
+            m_renderer = new DJVURenderer(source->path);
+            if(source->resolution == -1) source->resolution = 300;
+        }
+        else {
```
manisandro: Indentation looks wrong

qt/src/DisplayerToolSelect.cc
```diff
@@ -266,7 +268,7 @@ void NumberedDisplayerSelection::paint(QPainter *painter, const QStyleOptionGrap
 	painter->setRenderHint(QPainter::Antialiasing, false);
 	QRectF r = rect();
 	qreal w = 20. / m_tool->getDisplayer()->getCurrentScale();
-	w = qMin(w, qMin(r.width(), r.height()));
+	w = std::min(w, std::min(r.width(), r.height()));
```
manisandro: Is there actually a rationale for preferring std::{min,max} instead of qMin qMax or is it just personal preference?

qt/src/Kdjvu.hh
```diff
@@ -0,0 +1,301 @@
...
+        /**
+        * Return the page number of the page whose title is \p name.
+        */
+        int pageCount() const;
+    private:
+        class Private;
+        Private * const d;
+};
+
+#endif
```
manisandro: In the interest of keeping the codebase clean, I think this class can be heavily reduced, only a tiny fraction is actually used.

qt/src/DisplayerToolSelect.cc
ZaMaZaN4iK: Qt developers recommend use STL stuff instead of Qt, because we are writing on C++, STL becomes more and more powerful, std::min/max is default "min/max" for C++ developers. There are no reasons to use Qt's min/max.

qt/src/DisplayRenderer.hh
ZaMaZaN4iK: Is it problem for gImageReader? Do you want reduce compile time, right?

CMakeLists.txt
ZaMaZaN4iK: Yep, i agree.

qt/src/Kdjvu.hh
ZaMaZaN4iK: I thought about it. I am not sure which functionality will be used in the future. So i decided left all functionality.

Because if we will want to extend djvu support, in KDjvu wrapper we will be able to find needed functionality.

qt/src/DisplayRenderer.hh
manisandro: Well it's general good practice in c++ to only include things when you really need to. This particular instance will have a limited effect, but if you start doing it everywhere you will definitely notice. So easier to just always follow the good practice.

qt/src/Displayer.cc
ZaMaZaN4iK: Fixed

qt/src/DisplayRenderer.cc
ZaMaZaN4iK: Fixed

qt/src/DisplayRenderer.hh
ZaMaZaN4iK: Ok, I understood you point of view ;-)

manisandro:
Also, I don't know whether you were planning to do so after the review, but just in case: please try to keep the commits clean. The goal is to have one clean commit per feature. You can easily squash together many commits here which add/revert changes. Having a well-structured commit history helps a tone when browsing through the commits later on, be it to look for a regression, or to sync the gtk and qt codebases.

manisandro:
Don't know if you are familiar with `git rebase -i` ;)

ZaMaZaN4iK:
Yep i know about strategy "one feature - one commit". But sometimes i forgot about it. Sorry :-)

And yep, i worked before with git rebase :-)

ZaMaZaN4iK:
@manisandro Are there any more suggestions?

manisandro:
- The indentation still looks wrong in Displayer.cc
- Some `#include <algorithm>` (are they actually all needed? algorithm quickly gets indirectly included by  other headers) could be moved to the block with the other system includes (just keep them alphabetical) to reduce number of whitepsace changes
- The avgDeskew change isn't really necessary, but if you like the numbers fully typed out, fine with me, just make it a separate commit.
- Personally I'd still prefer to keep the djvu implementation taylored to what is actually needed, since I'll need to port it over to gtk later on. And since gImageReader isn't really indented to become an okular clone, it's safe to assume that much of the functionality there will never be needed.

ZaMaZaN4iK:
- Hm.. it's strange because i ran format script before commiting (astyle is installed). And i don't see a place where indentation is wrong.
- They are needed, because you can' rely on system headers structure inside. And if you will try to rely and sometime reshuffle/delete some header, code will be broken. So good practice is include what you use directly. It's not influence dramatically to compilation time. Moved all <algorithm> to right alphabetical place.
- Fixed.
- Fixed.

qt/src/Displayer.cc
```diff
@@ -107,7 +107,10 @@ bool Displayer::setCurrentPage(int page) {
 		if(source->path.endsWith(".pdf", Qt::CaseInsensitive)) {
 			m_renderer = new PDFRenderer(source->path);
 			if(source->resolution == -1) source->resolution = 300;
-		} else {
+        } else if(source->path.endsWith(".djvu", Qt::CaseInsensitive)) {
+            m_renderer = new DJVURenderer(source->path);
+            if(source->resolution == -1) source->resolution = 300;
+        } else {
 			m_renderer = new ImageRenderer(source->path);
```
manisandro: The indentation is wrong here.

qt/src/Displayer.cc
```diff
@@ -202,7 +205,11 @@ bool Displayer::setSources(QList<Source*> sources) {
 		DisplayRenderer* renderer;
 		if(source->path.endsWith(".pdf", Qt::CaseInsensitive)) {
 			renderer = new PDFRenderer(source->path);
-		} else {
+		}
+        else if(source->path.endsWith(".djvu", Qt::CaseInsensitive)) {
+            renderer = new DJVURenderer(source->path);
+        }
+        else {
 			renderer = new ImageRenderer(source->path);
```
manisandro: Also here wrong indentation

qt/src/Displayer.cc
ZaMaZaN4iK: Fixed.

manisandro:
I've cleaned up things and pushed the commits to the djvu branch

https://github.com/manisandro/gImageReader/tree/djvu

As general hints:
- Start you work with the against the latest master
- If somethings changes in master, do a git rebase master in your work branch
- Use git add -p ti selectively add relevant hunks to a commit
- If necessary, use git rebase -i to squash/reorder stuff
- Avoid whitespace changes
- You are free to force push in your work branch if you change things after the PR is created, that means you can without any problems rework previous commits, you don't necessarily need to add new commits for subsequent changes

Feel free to modify the commits to ensure you appear as author, and then things should be ready to merge to master!

ZaMaZaN4iK:
For me doesn't matter will i appear as author in a header of source or not. So you can merge it :-)

manisandro:
Merged, thanks a ton for your work!

PR #198, opened 2017-10-21
Add image processing to gImageReader

ZaMaZaN4iK:
Hello,

It's pull request about adding image processing stuff to gImageReader.

Pull request isn't completed yet, but here you can track developing process.


manisandro:
Should I already leave comments or is it still early work in progress?

ZaMaZaN4iK:
No, you should not review it yet. But you can suggest me the way, hot to return image fro, Image Processor to Displayer class. Because i didn't see any way to do it: i can get image FROM Displayer, but i can't set image.

manisandro:
This is actually non-trivial. The Displayer always queries the images directly from the source (i.e. an image file or pdf or djvu document) with the appropriate resolution, either for displaying in the gui, or for passing to tesseract for OCR. The getImage image method is only there for selection tools etc which need to operate on the currently displayed image. Conversely, a setImage method would make little sense because you would only be setting the image used for the UI at the current zoom level etc.

I see two options:
- Parametrize the processing steps, so that they are applied to each rendered image, as is done for brightness, contrast etc, see
https://github.com/manisandro/gImageReader/blob/master/qt/src/DisplayRenderer.cc#L31
But I fear this might be too slow.

- Otherwise you'd need to have the processing algorithms create a new temporary image, which is then passed to the displayer as a source. To make it elegant, you could actually extent the Source structure
https://github.com/manisandro/gImageReader/blob/master/qt/src/SourceManager.hh#L32
to allow specifying an alternative path to use instead of the path actually specified by the source, and if that path is not empty, then that image is used instead. This has the benefit that you don't need to pollute the source list in the UI with extra entries for the processing output. This alternative path entry would probably have to be a

    QMap<int, QString>

since for multi-page documents you will need a temporary image for each page. For multi-page there is the additional challenge when to generate these temproary images, since generating them all at once will probably be to slow. Probably better to just generate them on demand.

I hope this makes sense to you.

manisandro:
(You'll want to start from a fresh branch or rebase your old branch - merge is almost always something you **don't** want to do)

ZaMaZaN4iK:
@manisandro Can you build it and check it?

manisandro:
Please clean up your commits

PR #204, opened 2017-10-23
Choose PDF version during export

ZaMaZaN4iK:
Pull request about choosing PDF version during export.

It's not final version. I will ping you when all will be ready for merging.

### DIFF
--- a/qt/data/PdfExportDialog.ui
+++ b/qt/data/PdfExportDialog.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>411</width>
-    <height>436</height>
+    <height>496</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -190,7 +190,7 @@
       <property name="bottomMargin">
        <number>4</number>
       </property>
-      <item row="2" column="1">
+      <item row="3" column="1">
        <widget class="QSpinBox" name="spinBoxDpi">
         <property name="minimum">
          <number>50</number>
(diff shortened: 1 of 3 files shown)


ZaMaZaN4iK:
Hmm, very interesting: i don't see any difference between PDF 1.0 and PDF 1.7 internally, except PDF version in a header.

manisandro:
I don't really know that much about the PoDoFo internals, so I suppose one would need to have a look at the sourcecode to check what effect the version actually has.

ZaMaZaN4iK:
Ok. Looks like that we use something from any modern PDF standard and PoDoFo simply ignore version. Also will be good idea to ask about it in PoDoFo mailing list or any IRC channel.

ZaMaZaN4iK:
> On Wed, 2017-10-25 at 01:14 +0300, Alexander Zaitsev wrote:
> I create some document with different PDF versions and compared it 
> binary. What i saw: documents are identical inside, except PDF
> version in a PDF header.
>
>	Hi,
>as far as I can tell (and after grep of the code for GetPdfVersion()
>usages), there is not done any checking of the version when writing the
>output, it's left up to the user to not use primitives which cannot be
>used in the version the user chose.
>	Bye, zyx

Accoring to the letter, We should specify PDF writer for every PDF version. E.g. if you want to work with PDF 1.0 we can't use PDF streams, etc.

I think there are a lot of work here, so i don't know is it reasonable or not do it right now.

manisandro:
Abandoned for the time being.

PR #209, opened 2017-11-03
Fix couple of problems revealed by non-gcc compiler

bircoph:
First change checks for lcc compiler and its version and disables not yet supported C++ features the same way as for ARMEL.

Second change fixes OpenMP detection to use standard cmake find_package(openmp).


manisandro:
Thanks!

PR #212, opened 2017-11-05
Add feature: opening encrypted PDF

ZaMaZaN4iK:


qt/src/SourceManager.cc
```diff
@@ -44,6 +53,38 @@ QDataStream& operator>>(QDataStream& ds, Source*&) {
 	return ds;
 }
 
+Source::Source(const QString& _path, const QString& _displayname, bool _isTemp)
+				: path(_path), displayname(_displayname), isTemp(_isTemp) {
+	// Check is input PDF file encrypted
+	if(path.endsWith(".pdf")) {
```
manisandro: You'll want a Qt::CaseInsensitive here

qt/src/SourceManager.cc
```diff
@@ -44,6 +53,38 @@ QDataStream& operator>>(QDataStream& ds, Source*&) {
...
+			if (document->isLocked()) {
+				bool ok = false;
+				QString text;
+				while(true) {
+					text = QInputDialog::getText(MAIN, _("Enter password for the PDF"),
+														 _("Password:"), QLineEdit::Password,
+														 "", &ok);
+					document->unlock(text.toLocal8Bit(), text.toLocal8Bit());
+					if(!ok) {
+						throw std::invalid_argument("Encrypted PDF: skip file.");
```
manisandro: This block should probably appear right after the QInputDialog::getText?

qt/src/SourceManager.cc
```diff
@@ -44,6 +53,38 @@ QDataStream& operator>>(QDataStream& ds, Source*&) {
...
+					}
+					if(!document->isLocked())
```
manisandro: You could just use the return value of document->unlock

qt/src/SourceManager.cc
```diff
@@ -44,6 +53,38 @@ QDataStream& operator>>(QDataStream& ds, Source*&) {
...
+					{
+						break;
+					} else {
+						QMessageBox::information(MAIN, _("Open encrypted PDF"), _("Entered password is wrong."));
+					}
+				}
+				if (ok && !text.isEmpty()){
+					password = text.toLocal8Bit();
```
manisandro: You can just assign password before the break; above and avoid this code block altogether

qt/src/SourceManager.cc
ZaMaZaN4iK: fixed.

qt/src/SourceManager.cc
ZaMaZaN4iK: fixed.

qt/src/SourceManager.cc
ZaMaZaN4iK: fixed.

qt/src/SourceManager.cc
ZaMaZaN4iK: fixed

qt/src/SourceManager.cc
```diff
@@ -44,6 +54,34 @@ QDataStream& operator>>(QDataStream& ds, Source*&) {
...
+				bool ok = false;
+				QString text;
+				while(true) {
+					text = QInputDialog::getText(MAIN, _("Enter password for the PDF file"),
+														 _("Password:"), QLineEdit::Password,
+														 "", &ok);
+					if(!ok) {
+						throw std::invalid_argument("Locked PDF: skip file.");
+					}
+					if(!document->unlock(text.toLocal8Bit(), text.toLocal8Bit())) {
```
manisandro: Shouldn't that read `if(document->unlock(...))` (without the `!`)?

qt/src/SourceManager.cc
```diff
@@ -44,6 +54,34 @@ QDataStream& operator>>(QDataStream& ds, Source*&) {
 	return ds;
 }
 
+Source::Source(const QString& _path, const QString& _displayname, bool _isTemp)
+				: path(_path), displayname(_displayname), isTemp(_isTemp) {
+	// Check is input PDF file encrypted
+	if(path.endsWith(".pdf", Qt::CaseInsensitive)) {
+		std::unique_ptr<Poppler::Document> document(Poppler::Document::load(path));
+		if(document) {
+			if (document->isLocked()) {
```
manisandro: You could compress this to if(document && document->isLocked()).

qt/src/SourceManager.cc
```diff
@@ -44,6 +54,34 @@ QDataStream& operator>>(QDataStream& ds, Source*&) {
 	return ds;
 }
 
+Source::Source(const QString& _path, const QString& _displayname, bool _isTemp)
+				: path(_path), displayname(_displayname), isTemp(_isTemp) {
+	// Check is input PDF file encrypted
```
manisandro: // Check whether input PDF file is encrypted

qt/src/SourceManager.cc
ZaMaZaN4iK: i checked it - '!' is right here.

qt/src/SourceManager.cc
ZaMaZaN4iK: fixed

qt/src/SourceManager.cc
ZaMaZaN4iK: fixed.

qt/src/SourceManager.cc
manisandro: Odd, shouldn't `unlock` return `true` if unlocking was successful and `false` if not?

qt/src/SourceManager.cc
```diff
@@ -44,6 +54,32 @@ QDataStream& operator>>(QDataStream& ds, Source*&) {
...
+			bool ok = false;
+			QString text;
+			while(true) {
+				text = QInputDialog::getText(MAIN, _("Enter password for the PDF file"),
+											 _("Password:"), QLineEdit::Password,
+											 "", &ok);
+				if(!ok) {
+					throw std::invalid_argument("Locked PDF: skip file.");
+				}
+				if(!document->unlock(text.toLocal8Bit(), text.toLocal8Bit())) {
```
manisandro: I checked the poppler source code, and looks like the return value of `unlock` is whether the document is still locked. So `false` means that the document is not locked after the unlock attempt, hence success... So yes, ! is correct.

ZaMaZaN4iK: :+1:

qt/src/SourceManager.cc
```diff
@@ -44,6 +54,32 @@ QDataStream& operator>>(QDataStream& ds, Source*&) {
...
+	// Check whether input PDF file is encrypted
+	if(path.endsWith(".pdf", Qt::CaseInsensitive)) {
+		std::unique_ptr<Poppler::Document> document(Poppler::Document::load(path));
+		if(document && document->isLocked()) {
+			bool ok = false;
+			QString text;
+			while(true) {
+				text = QInputDialog::getText(MAIN, _("Enter password for the PDF file"),
+											 _("Password:"), QLineEdit::Password,
+											 "", &ok);
```
manisandro: Testing it I noticed that, since the dialog is by default very small, the the user most likely won't see the entire window title, but just the "Password:" text. I'd personally use "Enter password for the PDF file" as label and something simple like "Protected PDF" as title.

qt/src/SourceManager.cc
```diff
@@ -44,6 +54,32 @@ QDataStream& operator>>(QDataStream& ds, Source*&) {
...
+				if(!ok) {
+					throw std::invalid_argument("Locked PDF: skip file.");
+				}
+				if(!document->unlock(text.toLocal8Bit(), text.toLocal8Bit())) {
+					password = text.toLocal8Bit();
+					break;
+				} else {
+					QMessageBox::information(MAIN, _("Open encrypted PDF"), _("Entered password is wrong."));
+				}
```
manisandro: The double dialog is kinda annoying.. What do you think about just showing the password dialog again, so that users who mistyped can just continue typing the next attempt instead of having to click away the info dialog? Many other applications also just keep showing the password dialog until the input is correct or the user explicitly cancels.

ZaMaZaN4iK: Oh... i thought about it. But with additional MessageBox i give to user some feedback.

qt/src/SourceManager.cc

qt/src/SourceManager.cc
manisandro: You could pass the previously entered to the `QInputDialog::getText`, so that the user will see that the dialog is re-shown with the previous input selected, which is pretty standard behaviour and should make users realize that the input was wrong.

ZaMaZaN4iK: Ok, i disabled additional messagebox

qt/src/SourceManager.cc
```diff
@@ -44,6 +54,30 @@ QDataStream& operator>>(QDataStream& ds, Source*&) {
...
+	// Check whether input PDF file is encrypted
+	if(path.endsWith(".pdf", Qt::CaseInsensitive)) {
+		std::unique_ptr<Poppler::Document> document(Poppler::Document::load(path));
+		if(document && document->isLocked()) {
+			bool ok = false;
+			QString text;
+			while(true) {
+				text = QInputDialog::getText(MAIN, _("Protected PDF"),
+											 _("Enter password for the PDF file"), QLineEdit::Password,
+											 "", &ok);
```
manisandro: Enter password for the PDF file**:** ;)

manisandro:
Cool, can you squash all commits to, append the [Qt] prefix, and then feel free to merge!

qt/src/SourceManager.cc
ZaMaZaN4iK: I am not sure:
image
Is ':' required here?

manisandro: Ok this is probably debatable... Personally for strings that are not just titles, tooltips, menu entries or button labels I always terminate the string with some sort of punctuation (see also i.e. in the various UI files), so for consistency yes I'd add a colon here.

ZaMaZaN4iK: Ok :-)

Fixed.

ZaMaZaN4iK:
Check it and if all is fine, merge it into master.

manisandro:
Thanks!

PR #213, opened 2017-11-07
Feature: Save encrypted pdf

ZaMaZaN4iK:
Hello @manisandro 
Can you review it please?

### DIFF
--- a/qt/data/PdfExportDialog.ui
+++ b/qt/data/PdfExportDialog.ui
@@ -6,21 +6,14 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>411</width>
-    <height>451</height>
+    <width>629</width>
+    <height>585</height>
    </rect>
   </property>
   <property name="windowTitle">
    <string>PDF Export</string>
   </property>
   <layout class="QGridLayout" name="gridLayout">
-   <item row="13" column="0" colspan="2">
-    <widget class="QCheckBox" name="checkBoxPreview">
-     <property name="text">
-      <string>Show preview</string>
-     </property>
-    </widget>
-   </item>
    <item row="1" column="0" colspan="2">
     <widget class="QGroupBox" name="groupBoxImage">
      <property name="title">
@@ -125,7 +118,69 @@
      </layout>
     </widget>
    </item>
-   <item row="14" column="0" colspan="2">
+   <item row="3" column="0" colspan="2">
+    <widget class="QGroupBox" name="groupBoxEncryption">
+     <property name="title">
+      <string>Encryption (optional):</string>
+     </property>
+     <layout class="QGridLayout" name="gridLayout_5">
+      <item row="0" column="0">
+       <widget class="QLabel" name="labelPasswordOpen">
+        <property name="text">
+         <string>Password:</string>
+        </property>
+       </widget>
+      </item>
+      <item row="0" column="1">
+       <widget class="QLineEdit" name="lineEditPasswordOpen">
+        <property name="sizePolicy">
+         <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
+          <horstretch>0</horstretch>
+          <verstretch>0</verstretch>
+         </sizepolicy>
+        </property>
+        <property name="echoMode">
+         <enum>QLineEdit::Password</enum>
+        </property>
+        <property name="dragEnabled">
+         <bool>true</bool>
+        </property>
+       </widget>
+      </item>
+      <item row="1" column="0">
+       <widget class="QLabel" name="labelConfirmPasswordOpen">
+        <property name="text">
+         <string>Confirm password:</string>
+        </property>
+       </widget>
+      </item>
+      <item row="1" column="1">
+       <widget class="QLineEdit" name="lineEditConfirmPasswordOpen">
+        <property name="sizePolicy">
+         <sizepolicy hsizetype="Preferred" vsizetype="Fixed">
+          <horstretch>0</horstretch>
+          <verstretch>0</verstretch>
+         </sizepolicy>
+        </property>
+        <property name="echoMode">
+         <enum>QLineEdit::Password</enum>
+        </property>
+        <property name="dragEnabled">
+         <bool>true</bool>
+        </property>
+       </widget>
+      </item>
+     </layout>
+    </widget>
+   </item>
+   <item row="0" column="0">
+    <widget class="QLabel" name="labelOutputMode">
+     <property name="text">
+      <string>Output mode:</string>
+     </property>
+    </widget>
+   </item>
+   <item row="7" column="0" colspan="2">
     <widget class="QDialogButtonBox" name="buttonBox">
      <property name="orientation">
       <enum>Qt::Horizontal</enum>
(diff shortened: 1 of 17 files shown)


qt/src/SourceManager.cc
```diff
@@ -25,6 +25,7 @@
 #include <QFile>
 #include <QFileInfo>
 #include <QImageReader>
+#include <QListWidgetItem>
```
manisandro: I'm not really convinced it is worth the trouble to not rely on implicitly included headers, especially for situations like this were the implicit include is very unlikely to ever break. It's a lost battle to really consistently specify all necessary includes explicitly , and it just adds tons of include statements. As a maintainer of 150+ packages for Fedora, I very rarely get bitten by compilation failures because of internal header changes, and when I do, it is no big deal really. In my opinion the benefits of keeping the source files smaller outweigh the often theoretical problem of potential compilation failures in the future.

ZaMaZaN4iK: But we can get compilation erros in future during upgrading to a next Qt version. 
I understand your point of view, but i think that we should include what we use. If compilation time is critical for even in .cpp files (i agree with your police about includes in headers), i can delete this include.

But i recommend left it.

manisandro: Yeah well but when upgrading to the next Qt version, this is likely the least of issues. (I'd rather hope that by then C++ modules come along and solve the include mess for good.) But in all the projects I've been or am active in, I've never seen this enforced, simply because it is very hard to really enforce. It is not about compilation time, the overhead of the preprocessor discarding the already included header is probably not that big, but simply about practicality - and about avoiding 100 lines of includes in each file.

ZaMaZaN4iK: So... Must i remove it or not? :-)

manisandro: My preference is for minimizing the number of lines in the source files, so I'd remove it yes.

qt/data/PdfExportDialog.ui
```diff
@@ -170,103 +274,46 @@
...
-      </item>
-      <item row="1" column="0">
-       <widget class="QLabel" name="labelDithering">
-        <property name="enabled">
-         <bool>false</bool>
-        </property>
+       <widget class="QLabel" name="labelPasswordOpen">
         <property name="text">
-         <string>Dithering algorithm:</string>
+         <string>Password</string>
```
manisandro: For consistency with the remaining UI, please add a colon

ZaMaZaN4iK: Fixed.

qt/data/PdfExportDialog.ui
```diff
@@ -170,103 +274,46 @@
...
+      <item row="0" column="1">
+       <widget class="QLineEdit" name="lineEditPasswordOpen">
+        <property name="sizePolicy">
+         <sizepolicy hsizetype="Preferred" vsizetype="Preferred">
+          <horstretch>0</horstretch>
+          <verstretch>0</verstretch>
+         </sizepolicy>
+        </property>
+        <property name="echoMode">
+         <enum>QLineEdit::Password</enum>
```
manisandro: Most applications have a second "confirm" field. Up to you whether you want to include it.

qt/data/PdfExportDialog.ui
```diff
@@ -170,103 +274,46 @@
...
-       <widget class="QComboBox" name="comboBoxImageCompression"/>
-      </item>
+     <property name="maximumSize">
+      <size>
+       <width>393</width>
+       <height>16777215</height>
+      </size>
+     </property>
+     <property name="title">
+      <string>General</string>
```
manisandro: The other groupbox titles have a colon, either add it here, or we can remove it from the other titles (and update all translations... ;( )

manisandro:
Rest is looks good to me, when you're done please squash the commits as usual and rebase against master.

qt/data/PdfExportDialog.ui
ZaMaZaN4iK: Hm... Good idea! Thanks.

qt/data/PdfExportDialog.ui
ZaMaZaN4iK: Oh, again, one more colon :-)

qt/data/PdfExportDialog.ui
ZaMaZaN4iK: What should we do if password and confirm password are different? show messagebox with warning and after that show PDF export dialog again?

manisandro: I'd just set the background of the inputs to red as soon as the entries do not match and disable the export button. In general it's better to avoid to many dialogs popping up.

manisandro: FYI, the select pages dialog also uses the red background for error reporting, you can take inspiration from [1].

[1] https://github.com/manisandro/gImageReader/blob/master/qt/src/Recognizer.cc#L408

ZaMaZaN4iK: Ok, thank you. i will check way with red background.

ZaMaZaN4iK: Added support for red background. I have one trouble with disabling "Ok" button. As you know, we have in PDF export dialog QDialogButtonBox. And i don't know how i can disable only "Ok" button. Can you suggest me anything, except using two buttons instead of QDialogButtonBox?

ZaMaZaN4iK:
Also i ran astyle script for sources. you can review it :-)

qt/data/PdfExportDialog.ui
manisandro: You can use

    QDialogButtonBox::button(QDialogButtonBox::Ok)->setEnabled(false)

By the way how about only setting the background red when:
- The user has finished editing the confirm line edit and it does not match with the primary line edit
- The user has finished editing the primary line edit and the confirm line edit is not empty
(The ok button should be disabled immediately in any case as soon as the entries do not match).

This just to avoid that the user is confused because things turn red as soon as he/she starts typing.

manisandro: http://doc.qt.io/qt-5/qlineedit.html#editingFinished

ZaMaZaN4iK: Ok, i already implemeneted it. But you can check - it's not work wery well. And i prefer use way without editFinished() signal. And use only passwordChanged() signal + immediate disabling "Ok" button.

I hope, you will agree with me

manisandro: I'd suggest:

    connect(m_pdfExportDialogUi.lineEditPasswordOpen, SIGNAL(textChanged(QString)), this, SLOT(passwordChanged()));
    connect(m_pdfExportDialogUi.lineEditConfirmPasswordOpen, SIGNAL(textChanged(QString)), this, SLOT(passwordChanged()));
    
    /* ... */

    void OutputEditorHOCR::passwordChanged() {
      if(m_pdfExportDialogUi.lineEditPasswordOpen->text() == m_pdfExportDialogUi.lineEditConfirmPasswordOpen->text()){
        m_pdfExportDialogUi.lineEditConfirmPasswordOpen->setStyleSheet(QStringLiteral(""));
        m_pdfExportDialogUi.buttonBox->button(QDialogButtonBox::Ok)->setEnabled(true);
      } else {
        m_pdfExportDialogUi.lineEditConfirmPasswordOpen->setStyleSheet("background: #FF7777; color: #FFFFFF;");
        m_pdfExportDialogUi.buttonBox->button(QDialogButtonBox::Ok)->setEnabled(false);
      }
    }

ZaMaZaN4iK: Without additional slots for editingFinished() ?

ZaMaZaN4iK: Check current solution, pls.

manisandro:
Looks good, just one suggestion: I'd rename "General:" to "Encryption (optional):". This just to avoid confusion for users not sure what the role of the password is. (Also, maybe place the box at the end, after text settings).

Then remember to rebase the branch against the current master.

ZaMaZaN4iK:
@manisandro I merged from master into branch. So you can resolve all conflicts with this branch variant.

ZaMaZaN4iK:
@manisandro sorry, but i don't understand, how i can properly do rebase from master onto my branch after merging and squashing. If you will help me, it would be fine

PR #255, opened 2017-12-15
[Qt] Fix bug in HOCRAttributeEditor constructor

foghawk:
textChanged wasn't correctly triggering validateChanges.

Seems to have resolved some long hangs on ignoring words/adding words to dictionary. Haven't looked closely enough yet to be certain of why, but the typo's clear enough.


manisandro:
Good catch, thanks!

PR #258, opened 2017-12-17
Fresh image processing

ZaMaZaN4iK:
Hello @manisandro .

I have createdfresh branch for image processing stuff. Can you check it please?

### DIFF
--- a/qt/src/Displayer.hh
+++ b/qt/src/Displayer.hh
@@ -54,6 +54,7 @@ public:
 	double getCurrentScale() const {
 		return m_scale;
 	}
+	Source* getCurrentSource() const;
 	QString getCurrentImage(int& page) const;
 	QImage getImage(const QRectF& rect);
 	QRectF getSceneBoundingRect() const;
@@ -66,6 +67,7 @@ public:
 public slots:
 	void setAngle(double angle);
 	void setScaledImage(const QImage& image, double scale = 1.0);
+	bool renderImage();
 
 private:
 	enum class RotateMode { CurrentPage, AllPages } m_rotateMode;
@@ -123,7 +125,6 @@ private slots:
 	void queueRenderImage();
 	void scaleTimerElapsed();
 	void sendScaleRequest(const ScaleRequest& request);
-	bool renderImage();
 	void setRotateMode(QAction* action);
 	void rotate90();
 	void zoomIn() {
--- a/qt/src/Displayer.hh
+++ b/qt/src/Displayer.hh
@@ -65,6 +65,7 @@ public:
 
 public slots:
 	void setAngle(double angle);
+	void setScaledImage(const QImage& image, double scale = 1.0);
 
 private:
 	enum class RotateMode { CurrentPage, AllPages } m_rotateMode;
@@ -125,7 +126,6 @@ private slots:
 	bool renderImage();
 	void setRotateMode(QAction* action);
 	void rotate90();
-	void setScaledImage(const QImage& image, double scale);
 	void zoomIn() {
 		setZoom(Zoom::In);
 	}
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,5 @@
 build
+
+\.idea/
+
+cmake-build-debug/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,7 @@ SET(CXX_STANDARD_REQUIRED ON)
 
 
 # Dependencies
+FIND_PACKAGE(OpenCV REQUIRED)
 FIND_PACKAGE(Gettext REQUIRED)
 FIND_PACKAGE(PkgConfig REQUIRED)
 PKG_CHECK_MODULES(TESSERACT tesseract)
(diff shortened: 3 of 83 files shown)


qt/src/Displayer.hh
```diff
@@ -65,6 +65,7 @@ public:
 
 public slots:
 	void setAngle(double angle);
+	void setScaledImage(const QImage& image, double scale = 1.0);
```
manisandro: scaled image is just the image used in the UI, not the one used when recognizing. See `Displayer::getImage`.

ZaMaZaN4iK: It's just for testing. As far as i remember from our previous discuss about it, i should extend interface of Displayer.

manisandro: You'll probably want to do something similar to `DisplayRenderer::adjustImage`.

manisandro: Actually it depends on what operation you are performing, if they are operations which take a long time and should not be repeated everytime the image is redrawn, then you'll probably want to create a new temporary image with the operations applied, and then use that as source.

ZaMaZaN4iK: Hah, it's quite difficult to choose, because different image operations have different duration: from milliseconds to seconds. E.g. denoising is very expensive operation.

manisandro: So I'd work on the source level directly. Do you ever need the original image again after processing? If not, just clone the original image to a temporary location (see `SourceManager::savePixmap`), and then point `Source::path` to that location.

manisandro: Hmm though you'll need some special logic for multipage documents in that case.

ZaMaZaN4iK: Original image we need for adding "Cancel" (Ctrl+Z) for image processing algorithms. So we should store middleware images in RAM or on disk.

So isn't writing all images after every image operation huge overhead?

ZaMaZaN4iK:
Also please check UI. Now it's simple set of buttons for performing different algorithms.

qt/src/Displayer.hh
ZaMaZaN4iK: >Hmm though you'll need some special logic for multipage documents in that case.

Yep. Also i need specific logic for this kind of image processing: we have one image. And on one image we have more than one page and want to split it into different pages.

manisandro:
How do you envison the UI in the releasable version? If they remain simple button, I suppose a combobox with the respective entries in the advanced image controls toolbar would work just as well.

manisandro:
Also, how likely is that an algorithm will "wreck" an image? If that can happen, you'll either need a preview or a undo/redo. Preview is easier I suppose ;)

ZaMaZaN4iK:
>Also, how likely is that an algorithm will "wreck" an image? If that can happen, you'll either need a preview or a undo/redo. Preview is easier I suppose ;)

Of course, always there is some chance to make error in image processing algorithm :-) 

This question is very important and should be discussed. I see several ways to deal with it:
1) Every button apply algorithm on preview. And after some steps user can press "Apply" button and image from preview will replace source image.
2) You way with Ctrl+Z/Ctrl+Y. I think we also MUST implement it. It's very convinient way for program. I as user want to have this feature in gIR.
3) After every image processing algorithm ask to user about "Apply result from preview to source image?". I don't like this way :-)

ZaMaZaN4iK:
>How do you envison the UI in the releasable version? If they remain simple button, I suppose a combobox with the respective entries in the advanced image controls toolbar would work just as well.

I want to clone UI from Abbyy FineReader to gIR. And in FineReader they have separate dock widget for image processing.

Image proccesing UI will not be easy buttons. We will extend it with some customizable features: e.g. user must to be able to regulate denoise power, because we can't do it automatically.

manisandro:
For a start I'd just add a notification bar asking the user whether he wants to apply the change.

ZaMaZaN4iK:
Do you have any other warnings/suggestions?

manisandro:
Nope, with correct source logic and polished UI should be good!

ZaMaZaN4iK:
Ok, nice. Thank you for the review :-)

qt/src/Displayer.hh
ZaMaZaN4iK: @manisandro So what should i do for setting image to displayer in right way? Also i want to use preprocessed image in OCR. Can you explain me step by step? (i looked into adjustImage - no, we can't use it)

manisandro: I'd store the processed image as a temporary image and add it as a new source.

ZaMaZaN4iK: Where should i store this temporary image? What is the best way to add this temporary as a new source?

manisandro: See `SourceManager::savePixmap` for the temporary file logic, and then just do `SourceManager::addSource(tmpfilename)`.

ZaMaZaN4iK: btw, what do you think about using Doxygen for source documentation? For me gIR is new project and orientation inside for me is a little bit difficult. Can i use Doxygen for documenting source code, what i write?

manisandro: To be honest I don't know if method documentation helps a lot, since the method names are pretty clear and it is pretty pointless to write i.e.
    
    /* Set the sources
     * @sources The sources
     * @return True on success, false on failure
     */
    bool setSources(const std::vector<Source*> sources);

What one needs is a general overview of how the components play together, and I don't think that  doxygen helps much there.

ZaMaZaN4iK: You can look into image_processing sources: i use for it Doxygen.

manisandro: I'm familiar with doxygen, but again, I'm not convinced it helps much here besides blowing up the size of the source files.

ZaMaZaN4iK: About SourceManager::addSource method: but in this case as fasr as i understand my temporary item will be added to QListWidgetItem. Is it ok?

manisandro: Another idea could be to extend the `Source` struct by adding a `QMap<int,QString> preprocessedImages`, where you store, for each page, the preprocessed image. If such a page exists, the Displayer will render that image instead of the one of the original source. You can then give the user the option to revert to the original image, in which case the entry in `preprocessedImages` is simply discarded. This approach would also make it easy to apply the algorithms to all pages of a multipage document, without polluting the sources list with tons of entries.

ZaMaZaN4iK: I like this idea.

manisandro:
Hi @ZaMaZaN4iK, I've finally ended up releasing 3.3.0, so we can start again with feature work, if you are still interested in pursuing this.

manisandro:
Closing since this appears to be abbandoned.

PR #272, opened 2017-12-22
[Qt] feature: swap items

foghawk:
Allows reordering of sibling hOCR elements. Text being recognized out of order happens, and I needed this feature, but it's rare enough that this seemed sufficient.


foghawk:
You would know that better than I would—I'm learning Qt as I go. Actually, I went with this because it was sufficient for my purposes and IIRC you suggested drag & drop would be a pain ;)

If you want I can go back into the HOCRDocument API and see about breaking out a more general moveItem function to use as the underlying call.

manisandro:
Drag and drop is indeed always a pain, but sometimes it is the only way ;(

A generalized moveItem function would definitely be cool!

foghawk:
OK. I want to see about adding words first, but I'll come back to this afterwards.

foghawk:
Don't merge this yet, I'm not satisfied with the handling of page numbers.

The logical options are probably
1. renumber as appropriate iff all pages that would be affected are from the same source file, otherwise do nothing;
2. do nothing regardless.

The former gives reasonable results for the two main use cases (OCR of a single pdf without text and OCR of many separate page images), but is kind of dumb for anything in between, while the latter is kind of dumb regardless but at least has the virtue of consistency.

Is it, in fact, correct to restart page numbering for every source file when recognizing multiple files? (I'm not actually sure what's done with that information—if it's used for page labels in the generated PDF, I would argue that it isn't.)

manisandro:
gImageReader uses the `ppageno` attribute to know which page of the source document a particular hOCR page corresponds to (i.e. when you click on an item in the tree, gImageReader automatically opens the matching source file and shows the matching page). So the `ppageno` attribute should not be changed, it should always refer to the source page number.

foghawk:
Oh my god, of course it does. Fixed.

moveItem should be good to go, then. I bollixed up the merge so I'm going to open another pullreq for this on a clean branch.

PR #273, opened 2017-12-24
Output paper size support

ZaMaZaN4iK:
Hi!
This PR adds to gIR support to resize automatically pages to one format.


common/PaperSizeProcessor.cc
```diff
@@ -0,0 +1,128 @@
+/* -*- Mode: C++; indent-tabs-mode: t; c-basic-offset: 4; tab-width: 4 -*-  */
+/*
+ * Acquirer.cc
```
manisandro: Fix ;)

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -283,6 +286,18 @@ HOCRPdfExporter::HOCRPdfExporter(const HOCRDocument* hocrdocument, const HOCRPag
 	ui.comboBoxImageCompression->addItem(_("Jpeg (lossy)"), PDFSettings::CompressJpeg);
 	ui.comboBoxImageCompression->setCurrentIndex(-1);
 
+	ui.comboBoxPaperSizeMetric->addItem(_("mm"), static_cast<int>(PaperSizeMetrics::mm));
+	ui.comboBoxPaperSizeMetric->addItem(_("inch"), static_cast<int>(PaperSizeMetrics::inch));
+
+	ui.comboBoxPaperSize->addItem("Same as source", "Same as source");
+	ui.comboBoxPaperSize->addItem("Custom", "Custom");
```
manisandro: Label needs to localizable. Usually you don't use human strings as data.

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -449,7 +466,18 @@ bool HOCRPdfExporter::run(QString& filebasename) {
...
+
+	if(pdfSettings.paperSize == "Custom") {
+		ui.lineEditPaperSizeWidth->setEnabled(true);
+		ui.lineEditPaperSizeHeight->setEnabled(true);
+	}
+	if(pdfSettings.paperSize != "Same as source") {
+		Size<int> outputPaperSize = PaperSizeProcessor::getSizeMM(pdfSettings.paperSize.toStdString());
+		ui.lineEditPaperSizeWidth->setText(QString::number(outputPaperSize.width));
+		ui.lineEditPaperSizeHeight->setText(QString::number(outputPaperSize.height));
+	}
```
manisandro: Shouldn't this be done in paperSizeChanged?

common/PaperSizeProcessor.hh
```diff
@@ -0,0 +1,129 @@
...
+
+    T width, height;
+};
+
+enum class PaperSizeMetrics {
+    mm,
+    inch
+};
+
+enum class PaperSize {
```
manisandro: Where are these actually used?

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -283,6 +286,18 @@ HOCRPdfExporter::HOCRPdfExporter(const HOCRDocument* hocrdocument, const HOCRPag
 	ui.comboBoxImageCompression->addItem(_("Jpeg (lossy)"), PDFSettings::CompressJpeg);
 	ui.comboBoxImageCompression->setCurrentIndex(-1);
 
+	ui.comboBoxPaperSizeMetric->addItem(_("mm"), static_cast<int>(PaperSizeMetrics::mm));
+	ui.comboBoxPaperSizeMetric->addItem(_("inch"), static_cast<int>(PaperSizeMetrics::inch));
```
manisandro: I'd use the abbreviation for "inch", since you're also using "mm".

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -283,6 +286,18 @@ HOCRPdfExporter::HOCRPdfExporter(const HOCRDocument* hocrdocument, const HOCRPag
...
+
+	ui.comboBoxPaperSize->addItem("Same as source", "Same as source");
+	ui.comboBoxPaperSize->addItem("Custom", "Custom");
+	for(const auto& size : PaperSizeProcessor::paperSizes) {
+		ui.comboBoxPaperSize->addItem(size.first.c_str(), size.first.c_str());
```
manisandro: If the data is equal to the text, then saving the data is redundant.

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -684,3 +712,43 @@ QImage HOCRPdfExporter::getSelection(const QRect& bbox)
 {
 	return m_displayerTool->getSelection(bbox);
 }
+
+void HOCRPdfExporter::paperSizeChanged() {
+	delete ui.lineEditPaperSizeHeight->validator();
+	delete ui.lineEditPaperSizeWidth->validator();
```
manisandro: This looks pretty dangerous. Why delete the validator?

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -684,3 +712,43 @@ QImage HOCRPdfExporter::getSelection(const QRect& bbox)
...
+
+	if(ui.comboBoxPaperSizeMetric->itemText(ui.comboBoxPaperSizeMetric->currentIndex()) == "mm") {
```
manisandro: Since "mm" and "inch" are localized, this breaks easily if the translated strings differ. => Don't use human strings for logic, rather use `enum class PaperSizeMetrics`.

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -684,3 +712,43 @@ QImage HOCRPdfExporter::getSelection(const QRect& bbox)
...
+		ui.lineEditPaperSizeHeight->setValidator(new QIntValidator(0, 10000000));
+		ui.lineEditPaperSizeWidth->setValidator(new QIntValidator(0, 10000000));
+	} else if(ui.comboBoxPaperSizeMetric->itemText(ui.comboBoxPaperSizeMetric->currentIndex()) == "inch") {
+		ui.lineEditPaperSizeHeight->setValidator(new QDoubleValidator(0.0, 10000000.0, 2));
+		ui.lineEditPaperSizeWidth->setValidator(new QDoubleValidator(0.0, 10000000.0, 2));
+	}
+
```
manisandro: You can just adjust the params of the validator instead of creating a new one each time.

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -684,3 +712,43 @@ QImage HOCRPdfExporter::getSelection(const QRect& bbox)
...
+	auto textPaperSize = ui.comboBoxPaperSize->itemText(ui.comboBoxPaperSize->currentIndex());
+	if(textPaperSize == "Custom") {
+		ui.lineEditPaperSizeWidth->setEnabled(true);
+		ui.lineEditPaperSizeHeight->setEnabled(true);
+	} else if(textPaperSize == "Same as source") {
+		ui.lineEditPaperSizeWidth->setText("");
+		ui.lineEditPaperSizeHeight->setText("");
```
manisandro: I'd store width and height as a setting, and not clear it, even if "same as source" is selected. That way, a custom size the user specifies is kept persistent.

qt/src/hocr/HOCRPdfExporter.cc
ZaMaZaN4iK: Yep, it's just debug trash. Fixed.

qt/src/hocr/HOCRPdfExporter.cc
ZaMaZaN4iK: Fixed.

qt/src/hocr/HOCRPdfExporter.cc
ZaMaZaN4iK: What is the abbreviation for inch?

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -684,3 +712,43 @@ QImage HOCRPdfExporter::getSelection(const QRect& bbox)
...
+		ui.lineEditPaperSizeWidth->setEnabled(false);
+		ui.lineEditPaperSizeHeight->setEnabled(false);
+	} else {
+		ui.lineEditPaperSizeWidth->setEnabled(false);
+		ui.lineEditPaperSizeHeight->setEnabled(false);
+		if(ui.comboBoxPaperSizeMetric->itemText(ui.comboBoxPaperSizeMetric->currentIndex()) == "mm") {
```
manisandro: I'd use a PageSizeprocessor::getSize(PaperSizeMetrics unit) and move the size logic to there. Just store the PaperSizeMetrics as data in the comboBoxPaperSizeMetric, and you can call PageSizeprocessor::getSize directly with the data, and save some code.

qt/src/hocr/HOCRPdfExporter.cc
manisandro: in

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -684,3 +712,43 @@ QImage HOCRPdfExporter::getSelection(const QRect& bbox)
...
+			auto outputPaperSize = PaperSizeProcessor::getSizeMM(textPaperSize.toStdString());
+			ui.lineEditPaperSizeWidth->setText(QString::number(outputPaperSize.width));
+			ui.lineEditPaperSizeHeight->setText(QString::number(outputPaperSize.height));
+		} else if(ui.comboBoxPaperSizeMetric->itemText(ui.comboBoxPaperSizeMetric->currentIndex()) == "inch") {
+			auto outputPaperSize = PaperSizeProcessor::getSizeInch(textPaperSize.toStdString());
+			QLocale locale;
```
manisandro: This is a general question. currenty I've always used QString::number, which is not locale aware. If we want to use locale-aware number representation, then we should do so everywhere.

common/PaperSizeProcessor.cc
```diff
@@ -0,0 +1,128 @@
...
+
+        // North American paper sizes
+        {"Letter", Size<int>(216, 279)},
+        {"Legal", Size<int>(216, 356)},
+        {"Tabloid", Size<int>(279, 432)},
+        {"Ledger", Size<int>(432, 279)},
+        {"Junior Legal", Size<int>(127, 203)},
+        {"Half Letter/Memo", Size<int>(140, 216)},
+        {"Government Letter", Size<int>(203, 267)},
+        {"Government Legal", Size<int>(216, 330)},
```
manisandro: These are not localizable in this form. Though I don't know whether they should be localizable, since these are probably actual paper names that only exists in the respective language.

qt/src/hocr/HOCRPdfExporter.cc
ZaMaZaN4iK: I think we can support locale-dependent numbers here. And wiil be great, if we will extend this support to other parts of gIR

qt/src/hocr/HOCRPdfExporter.cc
ZaMaZaN4iK: Fixed

qt/src/hocr/HOCRPdfExporter.cc
ZaMaZaN4iK: Fixed.

qt/src/hocr/HOCRPdfExporter.cc
ZaMaZaN4iK: How can i do it, if in one case i must use QIntValidator, in another case - QDoubleValidator>

manisandro: Just `QDoubleValidator::setDecimals(0)`.

ZaMaZaN4iK:
BTW, what is the right way to set paper size in CreatePage function? Translate input size to inch and multiply on DPI?

manisandro:
I fear it is not that easy, I don't remember exactly what units PDF page units are, possibly points? The conversion from pixels to PDF units currently is hocrPageWidthPx * 72 / hocrPageDpi and analogous for height. So maybe
newhocrPageWidthPx = paperWidthIn * hocrPageDpi
newPdfPageWidth = newhocrPageWidthPx  * 72 / hocrPageDpi
=> newPdfPageWidth = paperWidthIn * 72

Try it out, I have I have no idea whether it is correct ;)

ZaMaZaN4iK:
What is 72? Magic DPI number?

manisandro:
Read https://en.wikipedia.org/wiki/Point_(typography)

manisandro:
Though since `1 point = 1/72 in`, if PDF units are indeed points, in theory it should be `newPdfPageWidth = paperWidthIn / 72`. Try out both, one of them should be correct ;)

qt/src/hocr/HOCRPdfExporter.cc
ZaMaZaN4iK: But in this case i should get validator from QLineEdit. QLineEdit returns only const QValidator* , so i can't change validator by this pointer.

manisandro: Just static-cast it, you know what type of validator the line edit has.

common/PaperSizeProcessor.cc
ZaMaZaN4iK: As fasr as i know we shouldn't localize it

common/PaperSizeProcessor.hh
ZaMaZaN4iK: When I wrote it, i thought it should be used in ComboBox. But now i am using string directly in combobox.

qt/src/hocr/HOCRPdfExporter.cc
ZaMaZaN4iK: I found one problem: if i set to QDoubleValidator setDecimals() to 0, QDoubleValidator allows to user enter single comma/dot without any decimal part. QIntValidator disallows.

manisandro: I don't really see a problem with that.

ZaMaZaN4iK: But QIntValidator is more convinient for "mm", and QDoubleValidator is more convinient for "inch" modes. Can you suggest more beautiful way deal with it? (except setDecimals() to 0)

manisandro: Just use cm for metric paper sizes, it is more common than mm.

ZaMaZaN4iK: Ok. Fixed.

ZaMaZaN4iK:
Ok, what else? :-)

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -283,6 +290,23 @@ HOCRPdfExporter::HOCRPdfExporter(const HOCRDocument* hocrdocument, const HOCRPag
...
+	ui.comboBoxPaperSize->addItem(_("Same as source"), "Same as source");
+	ui.comboBoxPaperSize->addItem(_("Custom"), "Custom");
+	for(const auto& size : PaperSizeProcessor::paperSizes) {
+		ui.comboBoxPaperSize->addItem(size.first.c_str(), size.first.c_str());
+	}
+
+	ui.lineEditPaperSizeHeight->setValidator(new QDoubleValidator(0.0, 10000000.0, 2));
+	ui.lineEditPaperSizeWidth->setValidator(new QDoubleValidator(0.0, 10000000.0, 2));
+
+	if(ui.comboBoxPaperSize->itemData(ui.comboBoxPaperSize->currentIndex()).toString() == "Same as source") {
```
manisandro: This isn't 100% correct, the fields should only be enabled in custom mode, and disabled in same as source and fixed-size-template modes. Also, I'd store it as a setting: add a setCurrentIndex(-1) here, to ensure no entry is selected, then add a ComboSetting below, which will then take care of restoring the previous choice, and the change of index triggered by the setting whill then call paperSizeChanged which will enable/disable the line edits as necessary.

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -309,6 +333,9 @@ HOCRPdfExporter::HOCRPdfExporter(const HOCRDocument* hocrdocument, const HOCRPag
 	connect(ui.lineEditPasswordOpen, SIGNAL(textChanged(const QString&)), this, SLOT(passwordChanged()));
 	connect(ui.lineEditConfirmPasswordOpen, SIGNAL(textChanged(const QString&)), this, SLOT(passwordChanged()));
 	connect(ui.checkBoxPreview, SIGNAL(toggled(bool)), this, SLOT(updatePreview()));
+	connect(ui.comboBoxPaperSize, SIGNAL(currentIndexChanged(int)), this, SLOT(paperSizeChanged()));
+	connect(ui.comboBoxPaperSizeMetric, SIGNAL(currentIndexChanged(int)), this, SLOT(paperSizeChanged()));
+    connect(ui.checkBoxLandscape, SIGNAL(toggled(bool)), this, SLOT(paperSizeChanged()));
```
manisandro: Wrong indentation

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -449,7 +476,11 @@ bool HOCRPdfExporter::run(QString& filebasename) {
 	pdfSettings.uniformizeLineSpacing = ui.checkBoxUniformizeSpacing->isChecked();
 	pdfSettings.preserveSpaceWidth = ui.spinBoxPreserve->value();
 	pdfSettings.overlay = ui.comboBoxOutputMode->currentIndex() == 1;
-	pdfSettings.detectedFontScaling = ui.spinFontScaling->value() / 100.;
+	pdfSettings.detectedFontScaling = ui.spinFontScaling->value() / 100.0;
+	pdfSettings.paperSize = ui.comboBoxPaperSize->itemData(ui.comboBoxPaperSize->currentIndex()).toString();
+
+	PaperSizeMetrics currentMetric = static_cast<PaperSizeMetrics>(ui.comboBoxPaperSizeMetric->itemData
+			(ui.comboBoxPaperSizeMetric->currentIndex()).toInt());
```
manisandro: It is not necessary to store the paperSize in the pdfSettings, since as far as I can see it is only used in this scope.

Just compute the inch size once here instead of doing it again and again for each page below.

qt/src/hocr/HOCRPdfExporter.cc
manisandro: I still think these should be stored as settings

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -684,3 +742,41 @@ QImage HOCRPdfExporter::getSelection(const QRect& bbox)
...
+
+void HOCRPdfExporter::paperSizeChanged() {
+	PaperSizeMetrics currentMetric = static_cast<PaperSizeMetrics>(ui.comboBoxPaperSizeMetric->itemData
+			(ui.comboBoxPaperSizeMetric->currentIndex()).toInt());
+
+	auto textPaperSize = ui.comboBoxPaperSize->itemData(ui.comboBoxPaperSize->currentIndex()).toString();
+	if(textPaperSize == "Custom") {
+		ui.lineEditPaperSizeWidth->setEnabled(true);
+		ui.lineEditPaperSizeHeight->setEnabled(true);
+	} else if(textPaperSize == "Same as source") {
```
manisandro: I'd not use human labels for logic (it is fragile, because you can mistake it for a UI label and changing it causes all sorts of troubles). Just use "source" and "custom".

common/PaperSizeProcessor.hh
```diff
@@ -0,0 +1,129 @@
...
+    RA2,
+    RA3,
+    RA4,
+    SRA0,
+    SRA1,
+    SRA2,
+    SRA4,
+    SRA3
+};
+
```
manisandro: Since this is unused I take it can be removed?

qt/src/hocr/HOCRPdfExporter.cc
ZaMaZaN4iK: Fixed.

common/PaperSizeProcessor.hh
ZaMaZaN4iK: It can. Fixed.

qt/src/hocr/HOCRPdfExporter.cc
ZaMaZaN4iK: Fixed.

qt/src/hocr/HOCRPdfExporter.cc
ZaMaZaN4iK: Fixed.

qt/src/hocr/HOCRPdfExporter.cc
ZaMaZaN4iK: Ok. I also prefer add to settings PaperSizeMetric combobox.

ZaMaZaN4iK: Fixed.

ZaMaZaN4iK:
Next iteration. What else? :-)

common/PaperSizeProcessor.cc
manisandro: Fix copyright headers

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -472,10 +521,17 @@ bool HOCRPdfExporter::run(QString& filebasename) {
...
-					PoDoFo::PdfPage* pdfpage = document->CreatePage(PoDoFo::PdfRect(0, 0, bbox.width() * docScale, bbox.height() * docScale));
-					pdfprinter.setPage(pdfpage, docScale);
+					PoDoFo::PdfPage* pdfpage;
+					if(paperSize == "source") {
+                        width = bbox.width() * docScale;
+                        height = bbox.height() * docScale;
+					}
+					pdfpage = document->CreatePage(PoDoFo::PdfRect(0, 0, width, height));
+                    double offsetX = 0.5 * (width - bbox.width() * docScale);
+                    double offsetY = 0.5 * (height - bbox.height() * docScale);
```
manisandro: wrong indentation (fix your editor ;))

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -684,3 +740,41 @@ QImage HOCRPdfExporter::getSelection(const QRect& bbox)
...
+	PaperSizeMetrics currentMetric = static_cast<PaperSizeMetrics>(ui.comboBoxPaperSizeMetric->itemData
+			(ui.comboBoxPaperSizeMetric->currentIndex()).toInt());
+
+	auto textPaperSize = ui.comboBoxPaperSize->itemData(ui.comboBoxPaperSize->currentIndex()).toString();
+	if(textPaperSize == "custom") {
+		ui.lineEditPaperSizeWidth->setEnabled(true);
+		ui.lineEditPaperSizeHeight->setEnabled(true);
+	} else if(textPaperSize == "source") {
+		ui.lineEditPaperSizeWidth->setText("");
+		ui.lineEditPaperSizeHeight->setText("");
```
manisandro: Doesn't this make the user potentially loose his previously entered custom width/height? I'd just hide the line edits when mode is "source".

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -684,3 +740,41 @@ QImage HOCRPdfExporter::getSelection(const QRect& bbox)
...
+		if(currentMetric == PaperSizeMetrics::cm) {
+			auto outputPaperSize = PaperSizeProcessor::getSizeMM(textPaperSize.toStdString());
+			widthText = locale.toString(static_cast<double>(outputPaperSize.width) / 10.0, 'f', 2);
+			heightText = locale.toString(static_cast<double>(outputPaperSize.height) / 10.0, 'f', 2);
+
+		} else if(currentMetric == PaperSizeMetrics::inch) {
+			auto outputPaperSize = PaperSizeProcessor::getSizeInch(textPaperSize.toStdString());
+			widthText = locale.toString(outputPaperSize.width, 'f', 2);
+			heightText = locale.toString(outputPaperSize.height, 'f', 2);
+		}
```
manisandro: You can save some code by just adding a `PaperSizeProcessor::getSize(unit)` function, moving the logic centrally to that function, instead needing the conditionals here and above in `OCRPdfExporter::run`.

qt/src/hocr/HOCRPdfExporter.cc
ZaMaZaN4iK: Just launched astyle. Fixed (sorry for this).

qt/src/hocr/HOCRPdfExporter.cc
ZaMaZaN4iK: How can i hide line edits? Which method?

manisandro: setVisible(false)?

ZaMaZaN4iK: Fixed.

qt/src/hocr/HOCRPdfExporter.cc
ZaMaZaN4iK: I think you meant these changes :-)

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -680,7 +732,45 @@ bool HOCRPdfExporter::setSource(const QString& sourceFile, int page, int dpi, do
...
+		ui.labelPaperSizeWidth->setVisible(false);
+		ui.lineEditPaperSizeWidth->setVisible(false);
+		ui.lineEditPaperSizeHeight->setVisible(false);
+	} else {
+		ui.labelPaperSizeHeight->setVisible(true);
+		ui.labelPaperSizeWidth->setVisible(true);
+		ui.lineEditPaperSizeWidth->setVisible(true);
+		ui.lineEditPaperSizeHeight->setVisible(true);
+		ui.lineEditPaperSizeWidth->setEnabled(false);
+		ui.lineEditPaperSizeHeight->setEnabled(false);
```
manisandro: You could pack all these widgets in a group (in the ui designer), then you can all these lines to just
   
    ui.groupCustomPaperSize.setEnabled(textPaperSize == "custom");
    ui.groupCustomPaperSize.setVisible(textPaperSize != "source");

ZaMaZaN4iK: I am not sure, is it required o r not. But i found that comboBoxPaperSizeMetric also should be invisible in "source" mode.

manisandro: Yep I'd agree that is should not be visible in source mode.

ZaMaZaN4iK: Not sure about landscape checkbox - maybe it also should be invisible in source mode.

manisandro: yes agree

ZaMaZaN4iK: Done.

ZaMaZaN4iK: Because i am not sure, how i should group these widgets. And i wouldn't create difficult group hierarchy.

manisandro: just use a widget with a gridlayout and put them all inside this widget

ZaMaZaN4iK: All - which of them? 
`ui.checkBoxLandscape
ui.comboBoxPaperSizeMetric
ui.labelPaperSizeHeight
ui.labelPaperSizeWidth
ui.lineEditPaperSizeWidth
ui.lineEditPaperSizeHeight
ui.lineEditPaperSizeWidth
ui.lineEditPaperSizeHeight`

manisandro: All except `ui.checkBoxLandscape` (and actually you could use a combobox [Portrait, Landscape] for the orientation, which is more in line with what all other applications do. I'd lay out as follows:
[ComboPaperSize][ComboPaperOrientation]
[Group Width/Height/Units]

manisandro:
@ZaMaZaN4iK GTK work is taking me a bit longer than expected, but I should be able to look at this before the years end.

ZaMaZaN4iK:
Ok, thank you :-)

manisandro:
@ZaMaZaN4iK I've cleaned up the UI and simplified the logic a bit, the only additional though is whether PaperSizeProcessor.{cc,hh,} should just be called PaperSize, containing one class PaperSize, the units and size being declared in the scope of that class. Besides honoring the practice of one class per source file, it also avoids having a "Size" declaration at global scope, which can easily cause symbol collisions.

ZaMaZaN4iK:
I built gIR after your changes and i see this:
image
image

I think it's wrong (i mean indentation from the right border.)

manisandro:
Applied as 837b0a9af820acd6bd391484c35357f411df76f7

PR #283, opened 2017-12-30
Add check for text layer in PDF file

ZaMaZaN4iK:
Hello,

here is simple check for text layer in PDF. Probably it should be extended.

### DIFF
--- a/qt/src/SourceManager.cc
+++ b/qt/src/SourceManager.cc
@@ -89,6 +89,7 @@ int SourceManager::addSources(const QStringList& files) {
 	QListWidgetItem* item = nullptr;
 	QStringList recentItems = ConfigSettings::get<VarSetting<QStringList>>("recentitems")->getValue();
 	int added = 0;
+	QStringList filesWithText;
 	for(const QString& filename : files) {
 		if(!QFile(filename).exists()) {
 			failed += "\n\t" + filename;
@@ -110,6 +111,12 @@ int SourceManager::addSources(const QStringList& files) {
 		if(!querySourcePassword(filename, password)) {
 			continue;
 		}
+
+		// Check text layer.
+		if(!checkTextLayer(filename)) {
+			filesWithText.push_back(filename);
+		}
+
 		Source* source = new Source(filename, QFileInfo(filename).fileName(), password);
 		item = new QListWidgetItem(QFileInfo(filename).fileName(), ui.listWidgetSources);
 		item->setToolTip(filename);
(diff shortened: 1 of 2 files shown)


ZaMaZaN4iK:
Also can be extended for DJVU files by using djvulibre.

manisandro:
Hi
As discussed, this will force the user to click "ok" multiple times if multiple documents are opened at once. So in this form I don't think it is a good approach.

ZaMaZaN4iK:
Which another ways do you suggest? Maybe we can add some marks for pages with text layer in UI.

ZaMaZaN4iK:
Trying to add text layer detection for DJVU documents, but have no success yet.

manisandro:
I suppose we can add icons to the source list. But dunno, seems like a lot of work for a pretty theoretical scenario.

ZaMaZaN4iK:
With checking text layer on DJVU i met some troubles - djvulibre API for this is too strange. So i think we should concentrate on PDF only for now.

ZaMaZaN4iK:
>I suppose we can add icons to the source list. But dunno, seems like a lot of work for a pretty theoretical scenario.

Can you suggest any way for adding this feature?

qt/src/SourceManager.cc
```diff
@@ -123,6 +126,14 @@ int SourceManager::addSources(const QStringList& files) {
 		recentItems.prepend(filename);
 		++added;
 	}
+	// Show files with text here:
+	if(!filesWithText.empty()) {
+		QString messageFiles;
+		for(const auto& file : filesWithText) {
+			messageFiles += file + "\n";
+		}
```
manisandro: Just use QStringList::join

qt/src/SourceManager.cc
```diff
@@ -156,6 +168,26 @@ bool SourceManager::querySourcePassword(const QString& filename, QByteArray& pas
...
+	if(filename.endsWith(".pdf", Qt::CaseInsensitive)) {
+		std::unique_ptr<Poppler::Document> document(Poppler::Document::load(filename));
+		bool withoutText = true;
+		if(document) {
+			const int pagesNbr = document->numPages();
+
+			for (int i = 0; i < pagesNbr; ++i) {
+				QString text = document->page(i)->text(QRectF());
+				if(!text.isEmpty()) {
+					withoutText = false;
```
manisandro: There is no need to continue searching subsequent pages once you hit a page with text. Just return false here, and you can save yourself some lines of code since you don't need "bool withoutText" at all. (Note: shorter code is always a plus!)

ZaMaZaN4iK:
Are there any more suggestions/warnings?

manisandro:
There are conflicts which prevent merging

ZaMaZaN4iK:
Should i resolve it on my side or you can resolve it during merging?

manisandro:
Would be nice if you do it on your side. Just git fetch && git rebase origin/master

ZaMaZaN4iK:
Done. Now there are no conflicts.

manisandro:
Thanks!

PR #284, opened 2018-01-02
[gtk] remove unrequired pixdata conversions

ramcq:
gdk-pixbuf-pixdata is deprecated and GResource files can embed .png files directly without conversion


manisandro:
Cool, thanks!

PR #289, opened 2018-01-08
Add support to more fonts.

napasa:
Hi,  Because Pofodo can not support fonts I expected, So I add one more pdf export method which use QtPdf Module.


qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -779,3 +659,211 @@ void HOCRPdfExporter::updateValid() {
...
+    QPrinter printer;
+    printer.setOutputFormat(QPrinter::PdfFormat);
+    printer.setOutputFileName(outname);
+    printer.setResolution(m_hocrdocument->page(0)->resolution());
+    printer.setPageSize(pageSize);
+    printer.setCreator(ui.lineEditCreator->text());
+    printer.setFullPage(true);
+
+    QFont defaultFont = ui.checkBoxFontFamily->isChecked() ? ui.comboBoxFontFamily->currentFont() : ui.comboBoxFallbackFontFamily->currentFont();
+
```
manisandro: The code after this part should be pretty much printer-agnostic. Can't you keep common for both QPrinter and PoDoFo, and just have dedicated functions for the painter initialization?

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -779,3 +659,211 @@ void HOCRPdfExporter::updateValid() {
...
+        valid = false;
+    ui.groupBoxEncryption->setEnabled(valid);
+    ui.comboBoxImageCompression->setEnabled(valid);
+    ui.comboBoxImageFormat->setEnabled(valid);
+    ui.lineEditAuthor->setEnabled(valid);
+    ui.lineEditTitle->setEnabled(valid);
+    ui.lineEditProducer->setEnabled(valid);
+    ui.lineEditKeywords->setEnabled(valid);
+    ui.lineEditSubject->setEnabled(valid);
+    ui.lineEditKeywords->setEnabled(valid);
```
manisandro: You should be able to just disable the entire advanced tab, since none of the entries there is applicable to the QPainter export.

qt/src/hocr/HOCRPdfExporter.cc
napasa: Hi, What does dedicated functions mean?  It means same as special function?

napasa: Do you mean that I should add more error-check code in **exporterbyQt** function?

qt/src/hocr/HOCRPdfExporter.cc
napasa: QPainter supports **Creator** option

manisandro: Ah ok

qt/src/hocr/HOCRPdfExporter.cc
manisandro: No, I mean that you should be able to leave the entire block of code below in `HOCRPdfExporter::run`, and just add something like `HOCRPdfExporter::createQPrinter` and `HOCRPdfExporter::createPoDoFoPrinter` which does the backend specific initialization of the printers.

napasa: I got your idea

napasa: Thanks  for your advice~

manisandro:
Hi @napasa are you still working on this?

manisandro:
I've done it myself in 8177ff37b4049327c500ea7361340069344d8e92, testing appreciated.

PR #294, opened 2018-01-13
[Qt] feature: swap items (now with arbitrary move)

foghawk:
Cleans up #272.

Tested, but only in the context of swap (i.e., among siblings). moveItem doesn't currently enforce any class relationships (since based on the merge and swap precedents that seems to be the caller's job); the only thing it will refuse to do with valid indices is try to move an item into its own descendants.


qt/src/hocr/HOCRDocument.cc
```diff
@@ -82,6 +82,59 @@ bool HOCRDocument::editItemAttribute(const QModelIndex& index, const QString& na
 	return true;
 }
 
+// Might be more logical to accept oldParent and oldRow as parameters rather than itemIndex
+// (since we need them anyway), if all our callers are likely to know them. Swap does; drag&drop might not.
+QModelIndex HOCRDocument::moveItem(const QModelIndex& itemIndex, const QModelIndex& newParent, int newRow, bool page) {
```
manisandro: You can detect whether it is a page via `item->itemClass()`

qt/src/hocr/HOCRDocument.cc
```diff
@@ -529,6 +587,13 @@ void HOCRItem::addChild(HOCRItem* child) {
 	child->m_index = m_childItems.size() - 1;
 }
 
+void HOCRItem::insertChild(HOCRItem* child, int i) {
+	m_domElement.insertBefore(child->m_domElement, m_domElement.childNodes().at(i));
+	m_childItems.insert(i, child);
+	child->m_parentItem = this;
+	child->m_pageItem = m_pageItem;
```
manisandro: You'll also need to reassign the item indices (`HOCRItem::m_index`).

manisandro:
Thanks!  - I'm currently working on drag and drop in the HOCRTree, for which I'll gladly use your moveItem work. Do you still want an explicit swapItem method in the context menu in addition to drag and drop?

qt/src/hocr/HOCRDocument.cc
foghawk: Thanks. Done.

foghawk:
If you have some reason to want it gone, feel free to remove it. If not, I figure it can't hurt and might be more convenient occasionally (since drag-and-drop is more sensitive to user error).

manisandro:
No particular reason, just asking.

qt/src/hocr/HOCRDocument.cc
```diff
@@ -82,6 +82,54 @@ bool HOCRDocument::editItemAttribute(const QModelIndex& index, const QString& na
...
+		endRemoveRows();
+		beginInsertRows(newParent, newRow, newRow);
+		parentItem->insertChild(item, newRow);
+		endInsertRows();
+	} else if(item->itemClass() == "ocr_page") {
+		beginRemoveRows(QModelIndex(), oldRow, oldRow);
+		HOCRPage* pageItem = m_pages.takeAt(oldRow);
+		endRemoveRows();
+		beginInsertRows(QModelIndex(), newRow, newRow);
+		m_pages.insert(newRow, pageItem);
```
manisandro: This (and `m_pages.takeAt` above) won't re-assign the page indices. You should be able to just use `takeItem` instead of `m_pages.takeAt`, and if you add a generic `HOCRDocument::insertItem` you can actually use the same code regardless of whether the item is a page or not. The only page-item specific part remains the `emit dataChanged`.

foghawk: Ah, so it doesn't. Fixed, and insertItem (private, analogous to take) added. I believe that as insert and take both emit their own dataChanged for pages, it actually isn't necessary for moveItem to do so—the only page logic left is checking whether an invalid parent is OK.

Am I correct in assuming that hOCR id attributes are subordinate to the tree structure in determining exported content flow, and do not need to be updated?

manisandro: Id's are just there to be unique but don't play any other role. No need to change them.

manisandro:
As I see things should be good to merge, shall I?

foghawk:
Yes, go ahead!

manisandro:
I've manually merged this. In future please use `git rebase upstream/master` instead of `git merge` to pick up the latest changes from the upstream branch, as otherwise the PR becomes impossible to merge. You might also want to setup your identity in git.

Thanks for your work!

PR #295, opened 2018-01-13
Feature/pdf version new

ZaMaZaN4iK:
Hello @manisandro 
It's PR for adding possibility for choosing output PDF version.

### DIFF
--- a/qt/src/SourceManager.hh
+++ b/qt/src/SourceManager.hh
@@ -48,6 +48,9 @@ struct Source {
 
 	//Additional info from original file
 	QString author, title, creator, producer, keywords, subject;
+
+	//Pdf version
+	int pdfVersionMajor = -1, pdfVersionMinor = -1;
 };
 
 class SourceManager : public QObject {
--- a/qt/src/hocr/HOCRPdfExporter.cc
+++ b/qt/src/hocr/HOCRPdfExporter.cc
@@ -306,6 +306,17 @@ HOCRPdfExporter::HOCRPdfExporter(const HOCRDocument* hocrdocument, const HOCRPag
 	ui.lineEditPaperHeight->setValidator(new QDoubleValidator(0.0, 10000000.0, 1));
 	ui.lineEditPaperWidth->setValidator(new QDoubleValidator(0.0, 10000000.0, 1));
 
+	ui.comboBoxPdfVersion->addItem("1.0", PoDoFo::EPdfVersion::ePdfVersion_1_0);
+	ui.comboBoxPdfVersion->addItem("1.1", PoDoFo::EPdfVersion::ePdfVersion_1_1);
+	ui.comboBoxPdfVersion->addItem("1.2", PoDoFo::EPdfVersion::ePdfVersion_1_2);
+	ui.comboBoxPdfVersion->addItem("1.3", PoDoFo::EPdfVersion::ePdfVersion_1_3);
+	ui.comboBoxPdfVersion->addItem("1.4", PoDoFo::EPdfVersion::ePdfVersion_1_4);
+	ui.comboBoxPdfVersion->addItem("1.5", PoDoFo::EPdfVersion::ePdfVersion_1_5);
+	ui.comboBoxPdfVersion->addItem("1.6", PoDoFo::EPdfVersion::ePdfVersion_1_6);
+	ui.comboBoxPdfVersion->addItem("1.7", PoDoFo::EPdfVersion::ePdfVersion_1_7);
+
+	ui.comboBoxPdfVersion->setCurrentIndex(-1);
+
 	connect(ui.comboBoxOutputMode, SIGNAL(currentIndexChanged(int)), this, SLOT(updatePreview()));
 	connect(ui.comboBoxImageFormat, SIGNAL(currentIndexChanged(int)), this, SLOT(updatePreview()));
 	connect(ui.comboBoxImageFormat, SIGNAL(currentIndexChanged(int)), this, SLOT(imageFormatChanged()));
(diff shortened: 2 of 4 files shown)


qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -306,6 +306,17 @@ HOCRPdfExporter::HOCRPdfExporter(const HOCRDocument* hocrdocument, const HOCRPag
...
 	ui.lineEditPaperWidth->setValidator(new QDoubleValidator(0.0, 10000000.0, 1));
 
+	ui.comboBoxPdfVersion->addItem(_("1.0"), PoDoFo::EPdfVersion::ePdfVersion_1_0);
+	ui.comboBoxPdfVersion->addItem(_("1.1"), PoDoFo::EPdfVersion::ePdfVersion_1_1);
+	ui.comboBoxPdfVersion->addItem(_("1.2"), PoDoFo::EPdfVersion::ePdfVersion_1_2);
+	ui.comboBoxPdfVersion->addItem(_("1.3"), PoDoFo::EPdfVersion::ePdfVersion_1_3);
+	ui.comboBoxPdfVersion->addItem(_("1.4"), PoDoFo::EPdfVersion::ePdfVersion_1_4);
+	ui.comboBoxPdfVersion->addItem(_("1.5"), PoDoFo::EPdfVersion::ePdfVersion_1_5);
+	ui.comboBoxPdfVersion->addItem(_("1.6"), PoDoFo::EPdfVersion::ePdfVersion_1_6);
+	ui.comboBoxPdfVersion->addItem(_("1.7"), PoDoFo::EPdfVersion::ePdfVersion_1_7);
```
manisandro: I don't think it makes sense to translate these ;)

ZaMaZaN4iK: Okay :-)

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -362,6 +373,7 @@ HOCRPdfExporter::HOCRPdfExporter(const HOCRDocument* hocrdocument, const HOCRPag
 	ADD_SETTING(LineEditSetting("pdfexportinfoauthor", ui.lineEditAuthor, PACKAGE_NAME));
 	ADD_SETTING(LineEditSetting("pdfexportinfoproducer", ui.lineEditProducer, PACKAGE_NAME));
 	ADD_SETTING(LineEditSetting("pdfexportinfocreator", ui.lineEditCreator, PACKAGE_NAME));
+	ADD_SETTING(ComboSetting("pdfexportpdfversion", ui.comboBoxPdfVersion, 7));
```
manisandro: http://doc.qt.io/qt-5/qcombobox.html#findData instead of hard-coding

qt/src/hocr/HOCRPdfExporter.cc
```diff
@@ -784,5 +797,34 @@ void HOCRPdfExporter::addInfoClicked() {
...
+		case 5:
+			sourcePdfVersion = PoDoFo::EPdfVersion::ePdfVersion_1_5;
+			break;
+		case 6:
+			sourcePdfVersion = PoDoFo::EPdfVersion::ePdfVersion_1_6;
+			break;
+		default:
+			sourcePdfVersion = PoDoFo::EPdfVersion::ePdfVersion_1_7;
+			break;
+		}
```
manisandro: What about `static_cast<PoDoFo::EPdfVersion>(PoDoFo::EPdfVersion::ePdfVersion_1_0 + source->pdfVersionMinor)`?

ZaMaZaN4iK: Yes, we can... in this case we will rely on internal representation of this enum. But ok

manisandro: The enum is part of the public API, so it is never supposed to change.

manisandro: You might want to add a fallback to 1.7 of major is >= 2 for future proofness also.

manisandro: And clearly this fallback is updated as PoDoFo picks up possible new versions, if we remember to do so. But if we forget to do so, at least it does not end up using version 1.0 or another ancient version.

qt/src/SourceManager.hh
```diff
@@ -48,6 +48,9 @@ struct Source {
 
 	//Additional info from original file
 	QString author, title, creator, producer, keywords, subject;
+
+	//Pdf version
+	int pdfVersionMajor, pdfVersionMinor;
```
manisandro: You should default-initialize these to zero or whatever, and check against this when using the values later on.

manisandro: Or perhaps even -1.

ZaMaZaN4iK: Fixed.

ZaMaZaN4iK: But i case with pdf i shouldn't check it because Poppler will return me version of pdf in these variables

manisandro: But you cannot be sure that the currently selected source is a PDF file.

ZaMaZaN4iK: Argh. I could check it with previous switch...

ZaMaZaN4iK: Ok, i will add check for this

manisandro: Well you can leave the static_cast, just add one additional condition if(major <= 0 or major >= 2) { version = 1.7; }

ZaMaZaN4iK: Fixed.

manisandro:
Thanks!

PR #296, opened 2018-01-14
Replace qSort by std::sort.

ZaMaZaN4iK:
Hello @manisandro 
qSort is deprecated in Qt (you can see it in qSort definition). I replaced it with std::sort as Qt recommends.


manisandro:
Thanks!

PR #304, opened 2018-02-01
[Qt] Improve error messages

ZaMaZaN4iK:
Hello @manisandro 

A little bit improved error messages for scannerSane.
Probably in implementation should be used Optional instead of defaultMessage.


manisandro:
Thanks

PR #316, opened 2018-02-15
Make GenericName in desktop file translatable

innir:
Hi,

from a [bug report](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=889633) in Debian I learned that a translatable GenericName is important for KDE Plasma.

Best,
Philip


manisandro:
Thanks! Perhaps missing something obvious, but how does the translated string from the po/mo file actually end up in the desktop file?

innir:
Ah sure, a

`intltool-merge --desktop-style po/ data/gimagereader.desktop.in data/gimagereader.desktop`

would be needed in CMakeLists.txt + fiddling with the interface type. I'll update the PR.

innir:
Should work now. It adds a build dependency on `intltool`.

CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ ENDIF()
 
 # Configure files
 CONFIGURE_FILE(data/gimagereader.appdata.xml.in data/gimagereader-${INTERFACE_TYPE}.appdata.xml @ONLY)
-CONFIGURE_FILE(data/gimagereader.desktop.in data/gimagereader-${INTERFACE_TYPE}.desktop @ONLY)
+CONFIGURE_FILE(data/gimagereader.desktop.in data/gimagereader-${INTERFACE_TYPE}.desktop.in @ONLY)
```
manisandro: This breaks if 

    IF(UNIX AND INTLTOOL_MERGE_EXECUTABLE)

is false below, since the `.in` suffix will remain and hence the final desktop file will be created.

innir: True, fixed with next commit.

manisandro: This still won't handle i.e. MINGW though. 

If there is no harm on running `intltool-merge` on an already processed file, you could omit the .in suffix for the output here and just use `add_custom_target` below to process the file in-place. Otherewise, I suppose you'll need to do something like

    IF(UNIX)
        CONFIGURE_FILE(data/gimagereader.desktop.in data/gimagereader-${INTERFACE_TYPE}.desktop.in @ONLY)
        # add translations to desktop file
        [...]
    ELSE()
        CONFIGURE_FILE(data/gimagereader.desktop.in data/gimagereader-${INTERFACE_TYPE}.desktop @ONLY)
    ENDIF()

innir: But is the desktop file needed for MINGW? My impression was not...

manisandro: Well, the file is installed independently of the platform, i.e.

    INSTALL(FILES ${CMAKE_CURRENT_BINARY_DIR}/data/gimagereader-${INTERFACE_TYPE}.desktop DESTINATION share/applications)

As for whether it is actually useful, there is i.e. a `desktop-file-utils` package for cygwin [1], so I'd say *maybe*. But what's more important is to not have half-processed files lying around in the build directory.

[1] https://cygwin.com/packages/x86_64/desktop-file-utils/

innir: Actually I think the whole section https://github.com/manisandro/gImageReader/blob/master/CMakeLists.txt#L28-L32 could just get a IF(UNIX) ... ENDIF() and everything should be fine.

manisandro: Well then you'd have to also put `IF(UNIX) ... ENDIF()` around some install directives. It would be easier to just use `add_custom_target` and modify the file in place, which I'd expect to work. The less platform specific blocks, the cleaner the file.

innir: using `intltool-merge` in place does not work, I'll go for your fist idea.

manisandro:
Thanks!

PR #319, opened 2018-03-14
[Qt] Fix PDF rights for documents without password

ZaMaZaN4iK:
Hello @manisandro .

Here is small fix for PDF documents without password.


manisandro:
Thanks!

PR #326, opened 2018-03-25
[Qt] Add option to split textlines, paragraphs, and text blocks in HOCR editor

foghawk:
Added this as I couldn't live without it anymore. (With merge and swap, any reordering is now technically possible, if not necessarily convenient.) Selecting a contiguous block of words, textlines, or paragraphs and then choosing "Split" from the context menu creates a new sibling adjacent to the parent and moves items in the block there.

Also fixes a bug in which removeItem did not correctly recompute ancestor bounding boxes.

I looked into extending the remove option to multiple items as well. Unfortunately, this resulted in irregular segfaults _after_ the return from showTreeWidgetContextMenu, and as I have never been able to get the crash backtrace to work I'm not sure what might be causing them.

### DIFF
--- a/qt/src/hocr/HOCRDocument.cc
+++ b/qt/src/hocr/HOCRDocument.cc
@@ -231,10 +231,11 @@ bool HOCRDocument::removeItem(const QModelIndex& index) {
 	if(!item) {
 		return false;
 	}
-	recomputeBBoxes(item->parent());
+	HOCRItem* parentItem = item->parent();
 	beginRemoveRows(index.parent(), index.row(), index.row());
 	deleteItem(item);
 	endRemoveRows();
+	recomputeBBoxes(parentItem);
 	return true;
 }
 
--- a/qt/src/hocr/HOCRDocument.cc
+++ b/qt/src/hocr/HOCRDocument.cc
@@ -84,7 +84,7 @@ bool HOCRDocument::editItemAttribute(const QModelIndex& index, const QString& na
 	}
 	emit itemAttributeChanged(index, name, value);
 	if(name == "title:bbox") {
-		recomputeParentBBoxes(item);
+		recomputeBBoxes(item->parent());
 	}
 	return true;
 }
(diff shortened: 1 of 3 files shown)


manisandro:
Thanks! - I'll review it this evening.

manisandro:
Looks good, thanks! About the multi-remove option: if you can post a patch somewhere, I give it a look.

manisandro:
What do you think about renaming "Split" to "Split from parent"? I think it makes it more intuitive what the function does.

foghawk:
I almost did, but I guess I'm too terse for my own good. Yes, go ahead.

I'll put the multi-remove up in a branch tomorrow. (Do backtraces work in your build?)

manisandro:
I mostly run it inside the debugger when developing (in qt-creator), so crashes are always caught by the debugger. But also normally I don't recall a crash without the crash handler popping up, but I actually very seldomly run the application outside the debugger.

manisandro:
Another idea, though I don't know if it is compatible with your particular use case: what if split actually split the parent container into [before split] [split] [after split], so i.e.

    line
      word1
      word2
      word3 <- split
      word4

becomes

    line
      word1
      word2
    line
      word3
    line
      word4

foghawk:
The crash handler does pop up—it just doesn't give any backtrace. I'll look into qt-creator. Patch at https://github.com/foghawk/gImageReader/commit/89a52f51eee0773b762096e81c7f5ddfe45ebbf8.

As for the three-way split, I thought about that too, but decided against it.
With the current behavior:
- Common case (items to be split at beginning/end of parent) is one action, and simple.
- Uncommon case where the user wants the current behavior is one action.
- Uncommon case where the user wants the three-way behavior is two actions (split/split).

With the three-way behavior:

- Common case is one action (but is slightly more complicated to implement).
- Uncommon case where the user wants the three-way behavior is one action.
- Uncommon case where the user wants the current behavior is _three_ actions (split/swap/merge).

foghawk:
That's still three actions: split/drag/merge. I have had to do that split-out-of-the-middle thing once or twice with some badly recognized tables of contents.

No, I was looking at `actionRemoveItem` and I didn't notice you'd already fetched the index :)

manisandro:
What I find most confusing with the current behaviour is

    line
      text1 <--- split
      text2
      text3

giving

    line
      text2
      text3
    line
        text1

I'd rather expect

    line
        text1
    line
      text2
      text3

Or am I missing something about the rationale of the feature?

PR #342, opened 2018-05-29
Update Chinese translations

timothytylee:
Fully updated all Chinese translations


manisandro:
Thanks and sorry for the late reply, completely missed the ticket!

PR #351, opened 2018-06-21
not add space character for CJK

napasa:


napasa:
Hi, Is it like this .
ヽ(✿ﾟ▽ﾟ)ノ

qt/src/Utils.cc
```diff
@@ -221,3 +221,21 @@ QString Utils::getSpellingLanguage(const QString& lang) {
...
+	std::vector<std::pair<int, int>> cjkWordRange{{0x2480, 0x303f}, {0x31c0, 0x9fff}
+		, {0xf900, 0xfaff}, {0xfe30, 0xfe4f}, {0x20000, 0x2fa1f}};
+	for(int i = 0; i < cjkWordRange.size(); i++) {
+		if(unicode < cjkWordRange[i].first) {
+			return true;
+		} else if(unicode >= cjkWordRange[i].first && unicode <= cjkWordRange[i].second) {
+			return false;
+		} else {
+			continue;
+		}
```
manisandro: This `else continue` block is redundant.

qt/src/Utils.hh
```diff
@@ -47,6 +47,8 @@ QByteArray download(QUrl url, QString& messages, int timeout = 60000);
 
 QString getSpellingLanguage(const QString& lang = QString());
 
+bool SpacedWord(const QString& text);
```
manisandro: Use `spacedWord` for camel case consistency.

manisandro:
So as I understand, you decide to add a space between to words if `Utils::SpacedWord` returns `true` for both the left and the right word. But `Utils::SpacedWord` only checks the first character of a word. Does this make sense? Isn't it better to determine whether to add a space based on the last character of the left word and the first character of the right word (in the case of left-to-right text)?

napasa:
you are a good teacher

PR #353, opened 2018-06-22
not add space character for CJK.

napasa:


manisandro:
It is not necessary to open a new PR every time, just push new commits to your branch, the PR will get updated automatically.

napasa:
Thanks.I try to rebase my new commit on top of your current master. and check whether the PR will get updated automatically.

napasa:
I finished my rebase housework.

manisandro:
Please open one PR per feature

napasa:
How to open one PR per feature

manisandro:
Make one branch per feature

napasa:
why github not a button to split PR

manisandro:
Because it would have to create two separate branches anyways to satisfy the git workflow, so you can just as well create the branches yourself.

napasa:
you are right

manisandro:
Could you answer the question from PR #351:

So as I understand, you decide to add a space between to words if `Utils::SpacedWord` returns true for both the left and the right word. But `Utils::SpacedWord` only checks the first character of a word. Does this make sense? Isn't it better to determine whether to add a space based on the last character of the left word and the first character of the right word (in the case of left-to-right text)?

napasa:
Yes, you are right. I have changed the detect method as you said, determined it based on the last character of the left-word and the first character of the right word.

YangtseSu:
There are still many uneeded spaces.

dogtopus:
What's the status of this PR? Looks like it hasn't been active for 3 years...

PR #354, opened 2018-06-27
resize content in proportion to source when Paper size is not same as source

napasa:


manisandro:
Please briefly explain the rationale

napasa:
For example you have export a pdf of which source paper size is equal to A4. Now you have to resize the paper size for rationale. Once in the circumstance like that,  if you enlarge paper size without enlarge font size in equal ratio you will  that recognized result is printed totally in up-left corner, In contrast you want to narrow your paper size, the content will be printed over boundary.

manisandro:
The idea of the paper size choice is not to re-layout the document, but simply to ensure that the output PDF has a consistent page size (considering that scanned images sizes often are cropped a bit more/less than the actual source page size). If you start scaling the font:
- You risk that every page has a slightly different text size, which looks bad
- You have to re-layout the elements on the page (i.e. recompute the bounding boxes). In particular, this breaks images, in the sense that the scaled text will likely flow into images.

As I see it, this is not a workable approach.

napasa:
I want to resize PDF paper size when some exported PDF paper size is very  unreasonable large. The upload pdf is  exported by gImageReader, Their paper size is 34.01 × 46.78 inch, and I want to narrow it to normal size. the follows file are the test file. 
[origin image](https://user-images.githubusercontent.com/8340899/42082532-cc98638a-7bba-11e8-8d7c-797c44bfd636.jpeg)

[output pdf](https://github.com/manisandro/gImageReader/files/2148500/webwxgetmsgimg.pdf)

napasa:
There is no possible that the original image I upload has a actual 34.01*46.78 inch paper size, It is just a common A4 scanned paper.  According to my understanding to wiki image resolution item, your method of converting pixel length to  point length only suit to pixel resolution is equal to spatial resolution. Actually the way I resize Paper Size is also wrong as you said. Maybe we should take consider the way Fedora document viewer  scales Paper Size.

manisandro:
If your source files are regular images without dpi information (as opposed to PDFs), it is in fact likely that the output size will be arbitrary. What one could do is add a checkbox "rescale contents to match paper size" if the chosen paper size is not "same as source". If checked, it would scale font sizes and bounding boxes accordingly.

manisandro:
HI, are you still planning to work on this?

PR #361, opened 2018-08-02
[Qt] don't allow Utils::documentsFolder() to supply nonexistent dirs

foghawk:
Qt5 guarantees that standard paths aren't empty, but not that they actually exist.

Most of this function's callers could already cope, but the crash handler tried to dump unsaved work in the returned location, which would totally fail.


manisandro:
Thanks!

PR #364, opened 2018-09-12
[Qt] [hOCR] Allow three-letter language codes (ISO 639-2 and -3)

foghawk:
I see now that this is analogous to your commit 84b6d82fd5d4e3ef248265aedd39227b526a49be (for GTK).

This is conservative and accepts only two- or three-letter language codes (based on ISO 639) and two-letter locales (based on my reading of ISO/IEC 15897), whereas 84b6d82fd5d4e3ef248265aedd39227b526a49be accepts 2+ letters for both. I don't think there are any actually extant identifiers on which they disagree, but I also don't know how this is supposed to work in Windows, so do whatever you think best.

(Not redundant with the Tesseract spec search as it handles scripts differently.)

### DIFF
--- a/qt/src/Utils.cc
+++ b/qt/src/Utils.cc
@@ -203,10 +203,7 @@ QByteArray Utils::download(QUrl url, QString& messages, int timeout) {
 
 QString Utils::getSpellingLanguage(const QString& lang) {
 	// If it is already a valid code, return it
-	if(lang.length() == 2) {
-		return lang;
-	}
-	if(QRegExp("^[a-z]{2}_[A-Z]{2}").indexIn(lang) != -1) {
+	if(QRegExp("[a-z]{2,3}(_[A-Z]{2})?").exactMatch(lang)) {
 		return lang;
 	}
 	// Treat the language as a tesseract lang spec and try to find a matching code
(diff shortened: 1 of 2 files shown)


foghawk:
Hm, none. Sorry, I thought it became relevant to the locale handling in general because of the request for the system's, but actually Qt abstracts this (and I should have been able to tell).

manisandro:
Ok let's leave it at 84b6d82 then.

foghawk:
That commit is GTK-only; it's still impossible to set or load three-letter codes in the Qt interface. If you prefer I can adjust this parsing to match that one.

manisandro:
There is also the Qt counterpart: https://github.com/manisandro/gImageReader/commit/2bf07f6334185f162953eb5cca388239210d3c2c#diff-af041c395075fdea339c16f7d9f7fdc0 (there should actually be no gtk- or qt-only features in the codebase up to the 3.3.0 tag)

foghawk:
_Ah._ I see the problem.

Both Qt and GTK allow three-letter codes in `Config::addLanguage` (via 84b6d82 and 2bf07f6 respectively), but neither allow them in `OutputEditorHOCR::attributeValidator`/`createAttrWidget` or `Utils::getSpellingLanguage`, which do independent validation. Although three-letter codes can be set in the config, they can't be entered during hOCR editing and will be overwritten if they exist in an hOCR file when loaded.

I've force-edited this branch to make both interfaces agree in all three locations. Again, sorry for the confusion.

manisandro:
Ah good catch, merged in https://github.com/manisandro/gImageReader/commit/544879bebc5e573506b64de4b540715ecbe6953f, thanks!

PR #365, opened 2018-09-12
[Qt] [hOCR] Propagate attributes down to manually added items 

foghawk:
When creating a manually added hOCR item, if an attribute is propagatable to the item's class and all of the item's relatives agree on its value, apply the consensus value to the new item as well. Very minor QoL improvement.


manisandro:
Looks good to me, thanks!

PR #366, opened 2018-09-12
[Qt] [hOCR] Don't auto-replace all dashes at the ends of lines

foghawk:
Sometimes they really are typeset like that. I corrected a number of them multiple times before I realized gIR was "fixing" them every time I opened the file.

I know character-accurate hyphenation is important for search, but as it is there's no way to opt out of this. (FWIW, I've never seen Tesseract in hOCR mode recognize an em-dash incorrectly or an en-dash at all.)

I've been thinking about advanced search (using tree structure and maybe attribute values along with text content), which would be handy for doing this more carefully as well as for some formatting issues, but it's not fully baked yet, sorry.

### DIFF
--- a/qt/src/hocr/HOCRDocument.cc
+++ b/qt/src/hocr/HOCRDocument.cc
@@ -620,11 +620,6 @@ HOCRItem::HOCRItem(const QDomElement& element, HOCRPage* page, HOCRItem* parent,
 		m_text = element.text();
 		m_bold = !element.elementsByTagName("strong").isEmpty();
 		m_italic = !element.elementsByTagName("em").isEmpty();
-		// For the last word items of the line, ensure the correct hyphen is used
-		QDomElement nextElement = element.nextSiblingElement();
-		if(nextElement.isNull()) {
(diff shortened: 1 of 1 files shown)


manisandro:
Perhaps this should be made opt-in? The reason the code is there is because someone had an issue with incorrect dashes resulting in non-searchable words in Acrobat, although perhaps tesseract has since changed behaviour and recognizes these correctly?

foghawk:
I agree that opt-in is best. (I didn't want to start looking at UI stuff yet and figured 'always off' was better than 'always on' in the interim, since that way manual editing is possible at least.)

Should it apply to opening hOCR files as well as recognizing? If so, such a setting should probably live in the preferences pane; otherwise, the 'Recognize' dropdown might be more helpful.

manisandro:
Since the only reason that logic was added was for PDF export, I suppose an idea would be to add the option to the PDF export dialog, and apply the substitution when actually writing the PDF in `HOCRPdfExporter::printChildren`, and not to the hOCR document itself.

PR #367, opened 2018-09-12
[Qt] [hOCR] Correctly trim some words containing non-word characters

foghawk:
This leads to handling contractions, hyphenated words, and acronyms/initialisms properly for the purposes of ignoring, adding, and suggesting.


manisandro:
Thanks!

PR #377, opened 2018-10-01
[Qt] [hOCR] Improve selection behavior after adding elements

foghawk:
Makes adding consistent with the rest of the actions that alter the tree. (In particular, it's now possible to add an item, select another item, and immediately swap them, where before the context menu would behave as though only the other element was selected.)


manisandro:
Thanks!

PR #378, opened 2018-10-01
[Qt] [hOCR] Correctly trim words followed by superscript numbers

foghawk:
The 'non-word' character class doesn't catch them since they're numbers. This isn't pretty, but I run into it a lot.


manisandro:
Thanks!

PR #379, opened 2018-10-01
[Qt] [hOCR] Rework spellcheck to handle unspaced dashes

### LINKED ISSUE #362: Spellcheck intelligently around unspaced em dashes
Many documents use the "word—word" convention for typesetting em dashes. Currently, gImageReader (at least in the Qt interface) interprets this as a single word, tries to spellcheck it as such, and inevitably fails, resulting in false-positive flags even when both words are correct.

I haven't tried to implement this yet because I'm uncertain about the details. Just suppressing the false positives would be simple enough, but how to handle suggestions? What to display if one word is valid and one is not? (Just splitting them would work in plain text mode, but doesn't seem realistic for hOCR.)

foghawk:
Resolves #362.

Changes `HOCRDocument` API slightly; `checkItemSpelling` is now public (taking an index) and should be used both for checking validity and (optionally) requesting suggestions.

### DIFF
--- a/qt/src/hocr/HOCRDocument.cc
+++ b/qt/src/hocr/HOCRDocument.cc
@@ -24,6 +24,7 @@
 #include <QSet>
 #include <QTextStream>
 #include <QtSpell.hpp>
+#include <cmath>
 
 #include "common.hh"
 #include "HOCRDocument.hh"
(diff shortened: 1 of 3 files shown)


manisandro:
Sorry for the delay - could you rebase your PR ontop of latest master? Then I'll merge.

foghawk:
Done.

manisandro:
Thanks!

PR #380, opened 2018-10-01
[Qt] [hOCR] Add option to specify DPI of image sources when exporting to PDF

### LINKED ISSUE #350: the pdf actual size of created pdf is too large
if I have a  3400*4677 pixels image, It will output 34.00 × 46.78 inch pdf which will cause Office Word can`t show word converted for that pdf normally.

foghawk:
Resolves #350.

The logic calculating output sizes coped poorly with image sources, since their "resolution" is actually the (percentage) scale factor used for recognizing. gIR would go ahead and interpret it as a physical ratio during export to PDF, leading to blowups in reported physical sizes, in image dimensions, and in file sizes if the user chose an accurate resolution. (For example, a a 7"×10" document scanned at 300 DPI into a 2100×3000 image file and recognized at 100% would, if exported at 300 DPI, result in a 21"×30" document containing a 6300×9000 image, for a serious inflation of the filesize.) It was possible to adjust the image dimensions by accounting for this behavior during the export, but not the physical size.

This adds an 'image source DPI' setting to the PDF export window, allowing gIR to perform image dimension _and_ DPI calculations accurately using a user-provided figure. (And arbitrary scaling with independent up/down-sampling, if you're into that. But I figured that the source and target DPI are what the user is most likely to have on hand.)

Edit: oh, also fixes #354 (albeit with different methodology. Actually, should this be allowed even when paper size is different from source? Might be handy to be able to both scale and crop/pad). I knew I'd seen more discussion of this around somewhere.

### DIFF
--- a/qt/data/PdfExportDialog.ui
+++ b/qt/data/PdfExportDialog.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>495</width>
-    <height>632</height>
+    <width>502</width>
+    <height>651</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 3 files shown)


manisandro:
Thanks!

PR #381, opened 2018-10-13
Fix some typos

stweil:
They were found using codespell.

Signed-off-by: Stefan Weil <sw@weilnetz.de>

### DIFF
--- a/qt/src/OutputTextEdit.cc
+++ b/qt/src/OutputTextEdit.cc
@@ -217,7 +217,7 @@ void OutputTextEdit::paintEvent(QPaintEvent* e) {
 			rect.setLeft(startLine.cursorToX(startLinePos) - 0.5);
 			painter.drawRect(rect);
 
-			// Draw selections inbetween
+			// Draw selections in between
 			QTextBlock block = startBlock;
 			int lineNo = startLine.lineNumber() + 1;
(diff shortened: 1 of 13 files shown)


qt/src/OutputTextEdit.cc
```diff
@@ -217,7 +217,7 @@ void OutputTextEdit::paintEvent(QPaintEvent* e) {
 			rect.setLeft(startLine.cursorToX(startLinePos) - 0.5);
 			painter.drawRect(rect);
 
-			// Draw selections inbetween
+			// Draw selections between
```
manisandro: This one was actually correct.

stweil: I'm not a native speaker, but all information which I can find says it's wrong (as does codespell). See for example https://www.grammarly.com/blog/in-between-or-inbetween/.

stweil: Do you want to keep it nevertheless?`Or replace it by in-between? Then I can update the commit.

manisandro: Ah right, should be *in between*.

stweil:
Updated the pull request to use _in between_ and added another commit which fixes spelling of a variable name.

manisandro:
Thanks!

PR #384, opened 2018-10-13
Fix some issues reported by LGTM

stweil:


stweil:
While the first commit is uncritical, the second one needs a closer review. I think that the "suggestions" from LGTM discovered code bugs.

manisandro:
Thanks! Could you apply these changes also to the Qt interface?

stweil:
The first commit is only valid for GTK. The second commit fixes GTK and Qt.

manisandro:
Ah missed the Qt change, ok perfect, thanks!

PR #393, opened 2018-12-26
Fix translation of list items in appdata

City-busz:
Each paragraphs and list items must be tagged, otherwise they are treated as English by appstream-generator.

This fixes the following error detected by `appstreamcli validate gimagereader-gtk.appdata.xml`:
```
E - gimagereader-gtk.appdata.xml:gimagereader-gtk.desktop:38
    The 'description' tag should not be localized in upstream metadata. Localize the 
    individual paragraphs instead.
```


manisandro:
Thanks!

Issue #396, opened 2019-01-13
Flatpak package


scx:
I created the flatpak package for gImageReader.
https://github.com/scx/gimagereader-flatpak (Gtk+3/Qt5)
https://github.com/scx/gimagereader-flatpak/tree/qt5 (Qt5)
https://github.com/scx/gimagereader-flatpak/tree/gtk3 (Gtk+3)

I am in the process of submitting it to the Flathub repo.
https://github.com/flathub/flathub/pull/810
https://github.com/scx/flathub/tree/com.github.manisandro.gImageReader-qt5

It would be nice if the author decided to participate in the further development of this flatpak package.

manisandro:
Thank you for working on this! I haven't yet looked in depth at flatpack, so currently I'm not really able to partecipate, but if there are specific issues, I'm happy to help as time allows.

scx:
The gtk variant is here:
https://github.com/flathub/flathub/pull/842
https://github.com/scx/flathub/tree/com.github.manisandro.gImageReader-gtk

salim-b:
> Do you know if the produced .metainfo.xml contains the correct version information?

I'm really not familiar with Flatpak internals. I tried to locate the appstream metadata of the installed gImageReader Flatpak and if I'm not mistaken it's found under `/var/lib/flatpak/app/io.github.manisandro.gImageReader/current/active/export/share/metainfo/io.github.manisandro.gImageReader.appdata.xml`, right?

That file contains the same outdated info as is found on FlatHub:

```sh
cat /var/lib/flatpak/app/io.github.manisandro.gImageReader/current/active/export/share/metainfo/io.github.manisandro.gImageReader.appdata.xml
```

<details><summary>Output</summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
<component type="desktop">
 <id>io.github.manisandro.gImageReader</id>
 <metadata_license>CC0-1.0</metadata_license>
 <project_license>GPL-3.0+</project_license>
 <name>gImageReader</name>
 <name xml:lang="cs_CZ">gImageReader</name>
 <summary>A graphical (gtk) frontend to tesseract-ocr</summary>
 <summary xml:lang="cs_CZ">Grafická (gtk) nadstavba pro engine tesseract-ocr</summary>
 <keywords>
 <keyword>ocr</keyword>
 <keyword>optical</keyword>
 <keyword>recognition</keyword>
 <keyword>tesseract</keyword>
 <keyword>pdf</keyword>
[...]

PR #397, opened 2019-01-13
Enhance AppData file

scx:
I would like to enhance the AppData file.

This is related to submitting my flatpak package to the Flathub repository.
https://github.com/manisandro/gImageReader/issues/396
https://github.com/flathub/flathub/pull/810

You should also consider changing [application id](https://github.com/scx/flathub/blob/030990011b8622d86a478a3314c33592b2539aae/gimagereader.appdata.xml.in#L4), as well as renaming AppData, [desktop](https://github.com/scx/flathub/blob/030990011b8622d86a478a3314c33592b2539aae/gimagereader.appdata.xml.in#L61) and icon file.
https://github.com/flathub/flathub/wiki/AppData-Guidelines#id
Although It is not required, it is highly recommended.

See also:
https://github.com/flathub/flathub/wiki/AppData-Guidelines
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-launchable
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-releases
https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-provides
https://hughsie.github.io/oars

### DIFF
--- a/data/gimagereader.appdata.xml.in
+++ b/data/gimagereader.appdata.xml.in
@@ -1,55 +1,68 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Copyright 2014 Sandro Mani <manisandro@gmail.com> -->
-<component>
- <id type="desktop">gimagereader-@INTERFACE_TYPE@.desktop</id>
- <metadata_license>CC0-1.0</metadata_license>
- <project_license>GPL-3.0+</project_license>
- <name>gImageReader</name>
- <name xml:lang="cs_CZ">gImageReader</name>
- <summary>A graphical (@INTERFACE_TYPE@) frontend to tesseract-ocr</summary>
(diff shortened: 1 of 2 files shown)


manisandro:
Could you use the same indentation as before for the appdata file so that what actually changed is easily recognizable?

scx:
@manisandro 

> Could you use the same indentation as before for the appdata file so that what actually changed is easily recognizable?

Sorry. Flathub maintainers tend to block pull requests due to trivialities such as "inappropriate" (in their opinion) indentation.

Convert single-space indent to double-space indent:
```
sed -i -re 's/^( +)/\1\1/' gimagereader.appdata.xml.in
```

Convert double-space indent to single-space indent:
```
sed -i -re 's/^( +)\1/\1/' gimagereader.appdata.xml.in
```

manisandro:
Thanks, I'll follow up with a whitespace-change commit to double-space indent then.

PR #425, opened 2019-11-16
Fix FTBFS with Tesseract 5.0-alpha

eighttails:
In Tesseract 5.0, TessBaseAPI class became non-copyable and gImageReader's initTesseract() fails to compile.
With this patch TessBaseAPI instance is handled by smart pointer.
This patch still keeps it compatible with Tesseract 4.0.


manisandro:
Thanks!

PR #463, opened 2020-06-03
Fix FTBFS with Tesseract 5-alpha

eighttails:
With Tesseract master now we have to include strngs.h explicitly to use STRING class.


manisandro:
Thanks!

PR #511, opened 2021-04-13
Support quazip 1.1

tgurr:
Short fix extending the current solution to add 1.1. In the long run it might probably make sense to just use `FIND_PACKAGE(QuaZip-Qt5)` instead of `FIND_PACKAGE(QuaZip5)`, like e.g. QMapShack does: https://github.com/Maproom/qmapshack/pull/350


manisandro:
Thanks!

Issue #473, opened 2020-09-17
Utilize GtkSourceView in the Output (in Plain Text mode)


AvtechScientific:
It would be nice to have [GtkSourceView](https://wiki.gnome.org/Projects/GtkSourceView) widget in the output instead of the standard GtkTextView. The widget substitution shouldn't be too problematic.

Advantages:
- source code syntax will be highlighted automatically. 
- documents editing in Markdown or other leighweight markup languages will be assisted by syntax highlighting.

manisandro:
Actually, it is already a GtkSourceView: https://github.com/manisandro/gImageReader/blob/master/gtk/data/OutputEditorText.ui#L384

I think you need to explicitly select the syntax highlighting language, it's not done automatically?

AvtechScientific:
> Actually, it is already a GtkSourceView: https://github.com/manisandro/gImageReader/blob/master/gtk/data/OutputEditorText.ui#L384

Hey, that's great!

> I think you need to explicitly select the syntax highlighting language, it's not done automatically?

True. [Gedit](https://wiki.gnome.org/Apps/Gedit) has a small drop-down menu on the bottom to select target language. Is it possible to add something like this?

![Select_Dar](https://user-images.githubusercontent.com/12374315/93525281-74bc2f80-f93e-11ea-9cff-0a87b6e44ea1.png)

manisandro:
Probably not any time soon, but I'm happy to accept pull requests.

AvtechScientific:
Hi,

there is somebody who wants to implement this functionality. She codes on Windows though. I looked into [info on compilation](https://github.com/manisandro/gImageReader/wiki/Compiling-gImageReader) but in its Windows section I only see the way to compile Qt GUI. Looks like the [Dockerfile](https://github.com/manisandro/gImageReader/blob/master/packaging/win32/Dockerfile) doesn't provide Gtk dependencies either. On the other hand [makeinstaller.sh](https://github.com/manisandro/gImageReader/blob/master/packaging/win32/makeinstaller.sh) mentions Gtk.

Could you, please, provide additional instructions/example on the above mentioned wiki page - how to compile Gtk GUI on Windows. Thank you!

manisandro:
Hi
I honestly have no idea whether the Gtk port works on Windows, I gave up on it at the time and switched to Qt for Windows. makeinstaller.sh still has the logic used at the time to produce the windows Gtk builds.

I'd recommend running Fedora in a virtual machine, where you can just run

[...]

Issue #516, opened 2021-05-10
Error during source code compilation


AvtechScientific:
Hi,

I would like to contribute to gImageReader/GTK and to this end I first try to compile current master branch code (commit 811132e25bd0b95a91ec23db39a438430461a7d0). I have installed dependencies as per wiki:

```
apt install debhelper cmake intltool libdjvulibre-dev libgtkspellmm-3.0-dev libgtkmm-3.0-dev libgtksourceviewmm-3.0-dev libcairomm-1.0-dev libpoppler-glib-dev libtesseract-dev libsane-dev libjson-glib-dev libxml++2.6-dev libpodofo-dev libzip-dev python3-gi uuid-dev
```

However I get this error, while trying to compile. Any ideas? Thank you!

```
~/gimage/gImageReader/build$ cmake -DINTERFACE_TYPE=gtk ..
-- GSettings schemas will be installed locally.
-- GSettings shemas will be compiled.
-- The C compiler identification is GNU 8.3.0
-- The CXX compiler identification is GNU 8.3.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- gtk interface will be built
-- Found Gettext: /usr/bin/msgmerge (found version "0.19.8.1") 
-- Found PkgConfig: /usr/bin/pkg-config (found version "0.29") 
-- Checking for module 'tesseract'
--   Found tesseract, version 4.0.0
-- Checking for module 'ddjvuapi'
--   Found ddjvuapi, version 3.5.27
-- Checking for module 'libpodofo'
--   Found libpodofo, version 
-- Checking for module 'sane-backends'
--   Found sane-backends, version 1.0.27
-- Checking for module 'enchant-2'
--   No package 'enchant-2' found
-- Checking for module 'enchant'
--   Found enchant, version 1.6.0
-- Checking for module 'gtkmm-3.0'
--   Found gtkmm-3.0, version 3.24.0
-- Checking for module 'gtksourceviewmm-4.0'
[...]

PR #520, opened 2021-05-24
closes issues #473, #516; adds ability to open files in OutputEditor

AvtechScientific:


gtk/src/OutputEditor.hh
```diff
@@ -74,4 +74,4 @@ public:
 	virtual void setLanguage(const Config::Lang& /*lang*/) {}
 };
 
-#endif // OUTPUTEDITOR_HH
+#endif // OUTPUTEDITOR_HH
```
manisandro, line 77: Can you configure your editor to not strip the trailing line breaks?

AvtechScientific: Hmm... I indeed used Gnome Builder for the first time to fix this issue. I must admit I'm not familiar with its UI. I could not find the setting that controls "stripping the trailing line breaks"... I tried to google it and here is what I've found:

https://gitlab.gnome.org/GNOME/meld/-/issues/134

"However, neither gedit nor Builder expose it as a visible preference..."

manisandro: In case just selectively pick the hunks you want to commit with `git add -p` and omit the whitespace changes from the commit.

AvtechScientific: I tried to use `git add -p`... I'm not sure whether it helped...

AvtechScientific: It looks OK now...

manisandro:
Thanks!

PR #523, opened 2021-06-05
Fix gettext package name

City-busz:
The two interface uses the same gettext package name as defined here:
https://github.com/manisandro/gImageReader/blob/master/CMakeLists.txt#L48


manisandro:
Thanks!

PR #526, opened 2021-06-20
Enables GTK UI cross-compilation for Windows (both 32 and 64 bit)

AvtechScientific:


AvtechScientific:
I didn't touch `.github/workflows/build.yml` though...

Another point: you incorporate release version into the file name of the `.exe` installers, e.g. `gImageReader_3.3.1_qt5_x86_64.exe`. This is redundant since release version is present in the versioned URL, i.e. `https://github.com/manisandro/gImageReader/releases/tag/v3.3.1` and prevents creation of static URLs that point on the latest version , like this: `https://github.com/manisandro/gImageReader/releases/latest/gImageReader_qt5_x86_64.exe`...

Is it possible not to mention the release version in the filenames? Or, alternatively, within the CI/CD create a soft link `gImageReader_qt5_x86_64.exe` that would point on `gImageReader_3.3.1_qt5_x86_64.exe`...

Thank you!

manisandro:
Thanks for your work and sorry for the slow reply (again...). Can you please rebase your commits?

As for CI/CD, I suppose I could just use "latest" as version.

AvtechScientific:
1. Done. I never did a rebase before - I hope I didn't mess up things in the process :-)
2. "latest" - is a great solution to the problem.
3. Don't forget to add the "gtk" build to the .github/workflows as you see fit...
Thank you!

packaging/win32/makeinstaller.sh
```diff
@@ -16,6 +16,9 @@ iface=${2:-qt5}
 # Note: This script is written to be used with the Fedora mingw environment
 MINGWROOT=/usr/$arch-w64-mingw32/sys-root/mingw
 
+export PKG_CONFIG_ALLOW_CROSS=1
+export PKG_CONFIG_PATH="$MINGWROOT/lib/pkgconfig:$MINGWROOT/"
+
```
manisandro, line 21: Curious, why do you need these?

AvtechScientific: I think I saw it in an example of cross-compiled gtk application... indeed it might be not needed... should I check without it and remove?

manisandro: I'ive cross compiled all sorts of stuff and never needed to tweak those, yes please try without

AvtechScientific: Yes, it compiles even without them...

manisandro:
Thanks!

manisandro:
@AvtechScientific I've enabled the Gtk CI builds and renamed the assets

PR #528, opened 2021-07-07
Better HiDPI support

eighttails:
Qt toolbar icons are displayed too small on 4K display.
This PR makes its appearance better.

Before:
image

After:
image


manisandro:
Thanks!

PR #533, opened 2021-07-18
Add Hebrew translation of the docs + fixes

BinyaminPekar:


manisandro:
Thanks and sorry for the late reply!

PR #563, opened 2022-01-24
[QT] Added "output pane" icon to Qt

will7007:
GTK has an icon for the output pane button, but Qt had no icon for the button at all, so I selected a similar-looking icon resource (pencil on a piece of paper) for consistency's sake. The issue is shown here, circled in red: gImageReader_noicon

The icon is visible in the sample screenshot in the readme, so it may have been accidentally removed at some point.

### DIFF
--- a/qt/data/MainWindow.ui
+++ b/qt/data/MainWindow.ui
@@ -782,8 +782,9 @@
     <bool>true</bool>
    </property>
    <property name="icon">
-    <iconset theme="document-edit">
-     <normaloff>.</normaloff>.</iconset>
+    <iconset theme="document-edit" resource="gimagereader.qrc">
+     <normaloff>:/extra-theme-icons/document-edit.png</normaloff>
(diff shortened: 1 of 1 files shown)


manisandro:
This shouldn't be required:
- extra-theme-icons is set as a fallback search path: https://github.com/manisandro/gImageReader/blob/8448a6a4078720cfa7179bc7fd452d5a473eabcf/qt/src/main.cc#L43
- document-edit.png exists as fallback icon: https://github.com/manisandro/gImageReader/blob/8448a6a4078720cfa7179bc7fd452d5a473eabcf/qt/data/extra-theme-icons/

Feel free to reopen if I'm missing something.

will7007:
Thanks for the help, I think this is a build issue on my end with Qt, assuming that it is building fine on your end. Despite `:/extra-theme-icons` being in the fallback search path, I can only get Qt Designer to find/display icon themes that are within my system's current icon theme (Yaru, Ubuntu 20.04.3 LTS). The built application doesn't contain the icon either, so it is not able to find the icon theme even after the fallback path is set.

PR #566, opened 2022-01-27
Bug fix: can't click on any item sometimes:

DonnKey:
For example, enable preview immediately after
recognizing a page: the whole page is selected and clicks do nothing.

### DIFF
--- a/qt/src/hocr/OutputEditorHOCR.cc
+++ b/qt/src/hocr/OutputEditorHOCR.cc
@@ -598,12 +598,13 @@ void OutputEditorHOCR::showItemProperties(const QModelIndex& index, const QModel
 		QRect minBBox;
 		if(currentItem->itemClass() == "ocr_page") {
 			minBBox = currentItem->bbox();
+			m_tool->clearSelection();
 		} else {
 			for(const HOCRItem* child : currentItem->children()) {
 				minBBox = minBBox.united(child->bbox());
 			}
+			m_tool->setSelection(currentItem->bbox(), minBBox);
 		}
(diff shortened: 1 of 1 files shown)


manisandro:
I don't see this as a bug: when the page item is selected, the respective bbox is show highlighted as when any other type of item is selected. You can just select any other item in the tree if you wish to select another item. It is true that you can't do so by clicking within the canvas, OTOH the same applies if you have a paragraph selected and want to select a word within that paragraph.

DonnKey:
I have a bias toward trying to enhance intuitiveness and discoverability
(that is, trying to make documentation as avoidable as possible.) It's
never perfect, but it can always be better.

This issue is a "you just have to know" sort of usability bug - the
workaround is not easily discoverable: the behavior is technically
reasonable (exactly as you describe), but the solution is one that requires
the sort of deep knowledge you (Sandro) have of the application to know
what to do.  For a naive user focused on the canvas as his interface,
suddenly everything stops working as he expects and nothing "obvious" gets
it out of that mode (not even esc). It took me a while to figure out what
was going on, and by now I know the app pretty well. (The workaround to use
the tree once I recognized the problem was obvious to me, but again I'm
also not sure a naive user would quickly see using that "secondary" (from
his point of view) interface as the solution.)

That said, when I get to the point of having easy to understand changes for
a better fix, we can discuss it more then. In the meantime, I can live with
it (as a developer) as it currently is now that I understand it. One of my
changes makes the likelihood of hitting the problem much worse, so we'll
have to deal with that at that time as well.

Donn

Issue #570, opened 2022-02-02
The proofread widget and Keyboard Shortcut windows are not showing up well


do2or:
The Proofreading widget is only displayed in the first word
and Keyboard Shortcut also has very small windows 
see the attached pictures 
![SHORTCUTS_WINDOW_SO_SMALL](https://user-images.githubusercontent.com/10603352/152139205-fab922bf-fac6-42fe-8da3-b879a1cdd00d.PNG)

![roofreading widget](https://user-images.githubusercontent.com/10603352/152139165-bd4c93e5-10f6-4579-aa16-f2141442b3ec.PNG)

AvtechScientific:
@do2or - you might want to take this [Rashi script model](https://gitlab.com/pninim.org/tessdata_heb_rashi) for this text. If you can provide a feedback on how it compares to ABBYY FineReader or any other OCR software - it would be great.

@manisandro - do you plan to add the proofreading widget to the GTK interface within your current effort or will it wait till next major release?

manisandro:
I've had one similar report about the shortcuts dialog being too small (also on Wndows). Can't reporduce it myself, but I guess I could just set a minimum size constraint for the dialog.

Regarding the incomplete widget, I guess it must have something to do with RTL text...

@AvtechScientific 
Regarding GTK: I have some code lying around, but I basically gave up, with GTK it seems pretty impossible to have an absolutely positioned floating widget within a parent widget, except if you go down all the way to overriding the painter level to draw the child widget into the parent widget, and then you also have to reimplement all the event handling for focus etc accordingly... The effort is just too large for the time I can invest. If you want, I can give upload the currrent state of the code, if you want to give it a go.

AvtechScientific:
@manisandro - I'm interested in the implementation of the [plain text proofreading](https://github.com/manisandro/gImageReader/issues/472). I'm not sure, but it sounds like in that approach there will not be the problem you've faced, as no overlapping of widgets should be required... But I think it is a good idea to upload your code in a separate branch anyway, as it can be used to borrow ideas common to both approaches.

manisandro:
Actually I see now that I already accidentally committed the files: https://github.com/manisandro/gImageReader/tree/master/gtk/src/hocr/wip_proofreadwidget

AvtechScientific:
P.S.: @manisandro - why haven't you released the GTK windows executables with 3.4.0? Even without the proof-reading widget the update is quite significant...

manisandro:
[...]

PR #577, opened 2022-02-23
Build HOCR shortcut dialog with QFormLayout

ferdnyc:
This does only the very, very basic work of replacing the HTML-table-constructed "Keyboard Shortcuts" dialog with one based on `QFormLayout`, fixing the dialog sizing/layout. 

The new code splits apart the keystrokes and descriptions, using `QKeySequence::toString(QKeySequence::NativeText)` to format the keystrokes themselves. Each description is now a separate translatable string (that's then inserted as a `QLabel` alongside the `QKeySequence::toString()` text), which should improve life for translators.

(`QKeySequence::NativeText` enables auto-translation of key names on all platforms, so those no longer need to be translated at all. Plus, on macOS  — if the documentation is to be believed (I don't have a  Mac to test it on) — it uses symbol-based formatting for the modifier keys, like you'd see next to a menu option if it has a shortcut.)

Because `QKeySequence` is incompatible with "merged" keystroke labels such as "`Ctrl+{Up,Down}`", I had to split those pairings into two separate keystrokes with identical descriptions. I feel the slightly less compact dialog is a worthwhile tradeoff, when stacked against the benefits from using Qt's auto-translations. Plus, I expect that redundancy to be temporary, as there is room for improvement in the current key bindings.

No change to any bindings, mappings, or descriptions is included here, that's for future PRs to tackle.

Partially addresses #570 

### Default English locale

Screenshot from 2022-02-23 08-20-02

### Executed as `LANG=de_DE gimagereader-qt6`

Screenshot from 2022-02-23 08-22-02

### DIFF
--- a/qt/src/hocr/HOCRProofReadWidget.cc
+++ b/qt/src/hocr/HOCRProofReadWidget.cc
@@ -112,6 +112,8 @@ private:
 	void keyPressEvent(QKeyEvent* ev) override {
 		HOCRDocument* document = static_cast<HOCRDocument*>(m_proofReadWidget->documentTree()->model());
 
+		#define KEY_IS_ARROW(a) (a == Qt::Key_Up || a == Qt::Key_Down || a == Qt::Key_Left || a == Qt::Key_Right)
+
 		bool nextLine = (ev->modifiers() == Qt::NoModifier && ev->key() == Qt::Key_Down) || (ev->key() == Qt::Key_Tab && m_wordItem == m_wordItem->parent()->children().last());
 		bool prevLine = (ev->modifiers() == Qt::NoModifier && ev->key() == Qt::Key_Up) || (ev->key() == Qt::Key_Backtab && m_wordItem == m_wordItem->parent()->children().first());
 		if(nextLine || prevLine) {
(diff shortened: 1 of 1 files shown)


manisandro:
Thanks! Regarding duplicate keybinding descriptions, actually I guess they could be de-duplicated by writing them like

    Adjust left bounding box edge to the left
    Adjust left bounding box edge to the right

which would make them also easier to understand.

ferdnyc:
@manisandro 

Well, actually, I was going to suggest a more radical departure. I suppose I might as well just lay it out here.

IMHO, rather than having to remember what combinations of keys do what, it would be simpler if there were one set of bindings to _move_ the bounding box, and a second set to adjust the _size_ of the bounding box (with the upper left corner being fixed.

IOW:
* <kbd>Ctrl</kbd>+(<kbd>Up</kbd>,<kbd>Down</kbd>,<kbd>Left</kbd>,<kbd>Right</kbd>): Change the position of the bounding box
* <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+(<kbd>Down</kbd>,<kbd>Right</kbd>): Expand the bounding box (vertically, horizontally)
* <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+(<kbd>Up</kbd>,<kbd>Left</kbd>): Shrink the bounding box (vertically, horizontally)

Yes, _certain_ things would take two keypresses where they used to only take one. (Though _move_ adjustments, which used to require two presses, would now only take one.) But on the other hand you never have to try to remember which side <kbd>Ctrl</kbd>+<kbd>Shift</kbd> adjusts.

ferdnyc:
I'll make the necessary adjustments!

Oh, BTW, I also already have the new shortcuts list (the old-new version, like in my screeenshots) processed into a `GtkShortcutWindow`-based `.ui` file, for the other side of the world.

(Amazingly, Gtk and Qt can't even agree on how to format _key combinations_.
* Qt: `QKeySequence("Ctrl+-")`
* GtkBuilder: `<property name="accelerator">&lt;Ctrl&gt;minus</property>`
 
So it took a fair amount of munging.)

manisandro:
Thanks for taking care of Gtk also!

(Sometimes it's an interesting exercise to stay up-to-date with both toolkits, somethings I wonder whether it's really worthwhile...)

ferdnyc:
Oh, the Gtk stuff is definitely a long way from anywhere; all I have is an un-integrated UI file. And since Glade doesn't even support `GtkShortcutWindow`, there's no way to preview or examine it without integrating it into the code first. Which I haven't even touched.

(Though I did create a branch [`ferdnyc:gtk-proofread`](https://github.com/ferdnyc/gImageReader/tree/gtk-proofread) in my fork, where I applied the contents of the `gtk/src/hocr/wip_proofreadwidget/` directory, since keeping that stuff on a topic branch instead just seemed like a win to me. If you want to create a branch in the project repo for it, I can open a PR targeting it and push that stuff here. The (updated) `ShortcutsDialog.ui` file is also on that branch.)

Anyway, the commit I just pushed _here_ implements the new keybindings in the Qt ui. Obligatory updated screenshot:

Screenshot from 2022-02-23 17-39-45

qt/src/hocr/HOCRProofReadWidget.cc
```diff
@@ -109,6 +112,8 @@ class HOCRProofReadWidget::LineEdit : public QLineEdit {
 	void keyPressEvent(QKeyEvent* ev) override {
 		HOCRDocument* document = static_cast<HOCRDocument*>(m_proofReadWidget->documentTree()->model());
 
+		#define KEY_IS_ARROW(a) (a == Qt::Key_Up || a == Qt::Key_Down || a == Qt::Key_Left || a == Qt::Key_Right)
```
ferdnyc, line 115: I fully accept whatever well-deserved mocking, derision, or wrinkled noses are thrown my way, as punishment for _this_ nastiness.

Also, I can take that out, if it's deemed too ugly to live. :disguised_face:

manisandro:
Thanks!

PR #582, opened 2022-03-04
Fix FTBFS in Windows with Qt6.

eighttails:
In only Windows NativeEventFilter is used.
QAbstractNativeEventFilter interface has changed in Qt6.


manisandro:
Thanks

PR #618, opened 2023-02-06
Update releases in appdata file

proletarius101:
To make the version information in https://flathub.org/apps/details/io.github.manisandro.gImageReader correct.

However, it might be a bit late as you have tagged your new version. It would be nice if you update the release version in appdata file in each tag.


manisandro:
Forgot about that tag... Fixed in https://github.com/manisandro/gImageReader/commit/2e2e5bc76a09164d86d42e203f51cab7729112c0

manisandro:
Thanks for the pointer!

PR #619, opened 2023-02-07
improve appstream metadata

proletarius101:
- add developer name (can be shown in https://flathub.org/apps/details/io.github.manisandro.gImageReader)
- set `Translate` link to the Weblate page rather than the ambiguous `README.md`


manisandro:
Thanks!

PR #624, opened 2023-03-22
hOCR editor: use "scan_res" instead of "res"

kba:
cf. https://github.com/kba/hocrjs/issues/68:

> The problem is the `res` property which is not part of the hOCR spec. I think this should be [`scan_res`](https://kba.github.io/hocr-spec/1.2/#scan_res), i.e. in https://github.com/manisandro/gImageReader/blob/8a046ec7bf64e996fbc81187306334fbd11873c9/qt/src/hocr/OutputEditorHOCR.cc#L268:
> 
> ```diff
> - 	attrs["res"] = QString::number(pageInfos.resolution);
> +	attrs["scan_res"] = QString::number(pageInfos.resolution);
> ```

ht @khashashin


manisandro:
Thanks

khashashin:
@kba @manisandro thank you for the fix 👍

PR #625, opened 2023-03-22
hOCR editor: quote "x_font" property

kba:
cf. https://github.com/kba/hocrjs/issues/67

The `x_font` attribute must be quoted in case it contains spaces, otherwise a parser won't know where the value ends.

Untested and unsure whether the attribute-reading code needs corresponding adaption to deal with the quotes.


manisandro:
Fixed in https://github.com/manisandro/gImageReader/commit/bfe810465f9a4e0e2c76e0a058884b123d46f3bd

PR #667, opened 2024-02-18
Make package release dates reproducible

innir:
Using `string(TIMESTAMP ...)` respects `SOURCE_DATE_EPOCH` since CMake 3.8, which makes it possible to fix the date used as release date in AppStream metadata and Fedora packaging.

I didn't change the PACKAGE_RFC_DATE as %z was anlyadded in CMake 3.26 and I didn't want to raise the required CMake version. Might be an option to just remove the in-tree Debian packaging finally?


manisandro:
Thanks - I guess the in-tree distro packaging can also be removed if there is a downside keeping it. Indeed also for Fedora and OpenSuse the specs are managed separately in the respective build environments.

bmwiedemann:
coming from openSUSE: For .spec files, it does not hurt, if they are also in git / tar. It can even help in packaging when updates require changes to the .spec file.

PR #672, opened 2024-05-12
Remove deprecated <comment> tag in the appstream metadata

proletarius101:
It doesn't exist in the [specification](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-screenshots) any more and it [prevents](https://buildbot.flathub.org/#/builders/19/builds/14658) the program from being built on flathub


manisandro:
Thanks!

manisandro:
I'd like to fix the enchant 2.7 issue first and also wait for tesseract-5.4.0 which is currently RC.

PR #689, opened 2024-11-19
fix: Failed to compile in Qt5 environment

Kakueeen:
Refactored the resource handling in CMakeLists.txt to support both Qt5 and Qt6 versions

Additionally, updated string splitting logic across multiple files to ensure compatibility with different Qt versions.

截图_选择区域_20241119094706
截图_选择区域_20241119095548


manisandro:
Which Qt5 version are you using?

Kakueeen:
> Which Qt5 version are you using?

5.11

manisandro:
Yes setting a minimum version can make sense!

PR #691, opened 2024-11-28
fix: [build] Add minimum Qt5 version requirement check

Kakueeen:
Add explicit version check to ensure Qt5 version is at least 5.15 when building with Qt5 interface. This prevents build failures with older Qt5 versions that may lack required features.

The check is added right after finding Qt5Concurrent package and will fail with a clear error message if an incompatible version is detected.


manisandro:
Thanks!

PR #693, opened 2024-12-09
fix: Fix memory leaks in Displayer and RecognitionMenu

Kakueeen:
Fix two memory management issues:
- Add missing deletion of m_imageItem in Displayer::setSources
- Set proper parent for PSM menu to ensure proper cleanup

These changes prevent memory leaks and ensure proper resource cleanup when the application exits or when components are destroyed.


manisandro:
Thanks!

PR #696, opened 2025-01-10
chore: remove some variable that never used.

GongHeng2017:
1. class member 'FileTreeModel::OutputFileRole' is never used.
2. Local variable 'pos' shadows outer varibale.
3. Variable 'tempPath' is assigned a value that is never used.


GongHeng2017:
@manisandro

PR #697, opened 2025-01-10
chore: some function argument name different.

GongHeng2017:
1. Variable 'child' can be declared as pointer to const.
2. Function 'moveItem' argument 3 names different: declaration 'row' definition 'newRow'.
3. Function 'swapItems' argument 2 names different: declaration 'startRow' definition 'firstRow'.
4. Function 'insertChild' argument 2 names different: declaration 'index' definition 'i'.


GongHeng2017:
@manisandro

manisandro:
Thanks, could you avoid the whitespace changes though?

GongHeng2017:
> Thanks, could you avoid the whitespace changes though?

You're welcome. I should have done it

PR #699, opened 2025-02-05
fix: Add null pointer check logic for pointer variables.

GongHeng2017:
1. Some array variable can be declared as const array.
2. Add null pointer check logic for pointer variables.


GongHeng2017:
@manisandro 
Hello, due to conflict issues, I resubmitted it.

manisandro:
Thanks

PR #706, opened 2025-03-04
fix(display): Replace invalid commas with semicolons in angle calcula…

JWWTSL:
…tions

- Fix syntax errors caused by Chinese commas in ternary operators.
- Ensure angle normalization works correctly during rotation.w


manisandro:
Thanks!

Issue #688, opened 2024-10-30
AppImage and macOS official package


efa:
are you open to PR for an official Linux AppImage and macOS package?

manisandro:
Sure!

manisandro:
Closing the issue, but feel free to submit the PR!

efa:
I made a script that generate the AppImage, but the resulting AppImage is impacted by:
https://github.com/manisandro/gImageReader/issues/707

efa:
new PR: https://github.com/manisandro/gImageReader/pull/728

Issue #707, opened 2025-03-05
Segmentation fault (core dumped)


efa:
I just build the GTK frontend on 2 different Debian 12 installations:

1) Debian12/WSL/Win10
```
$ sudo apt install  debhelper cmake intltool libdjvulibre-dev libgtkspellmm-3.0-dev libgtkmm-3.0-dev libgtksourceviewmm-3.0-dev libcairomm-1.0-dev libpoppler-glib-dev libtesseract-dev libsane-dev qtbase5-dev qttools5-dev qttools5-dev-tools libqtspell-qt5-dev libpoppler-qt5-dev libjson-glib-dev libxml++2.6-dev libpodofo-dev libquazip5-dev libzip-dev python3-gi uuid-dev
$ git clone https://github.com/manisandro/gImageReader.git
$ cd gImageReader
$ mkdir build
$ cd build
$ cmake -DINTERFACE_TYPE=gtk ..
$ make
$ export XDG_DATA_DIRS=$PWD:$XDG_DATA_DIRS
$ gImageReader/build$ gimagereader-gtk
(gimagereader-gtk:21354): Gtk-WARNING **: 17:53:23.208: Could not load a pixbuf from /org/gtk/libgtk/icons/16x16/status/window-close-symbolic.symbolic.png.
This may indicate that pixbuf loaders or the mime database could not be found.
**
Gtk:ERROR:../../../gtk/gtkiconhelper.c:495:ensure_surface_for_gicon: assertion failed (error == NULL): Failed to load /org/gtk/libgtk/icons/16x16/status/image-missing.png: Unrecognized image file format (gdk-pixbuf-error-quark, 3)
Bail out! Gtk:ERROR:../../../gtk/gtkiconhelper.c:495:ensure_surface_for_gicon: assertion failed (error == NULL): Failed to load /org/gtk/libgtk/icons/16x16/status/image-missing.png: Unrecognized image file format (gdk-pixbuf-error-quark, 3)
Segmentation fault (core dumped)
```

2) Debian12 (on real HW)
```
/gImageReader/build$ gimagereader-gtk
Segmentation fault
```

and a windows open with the backtrace:

```
gImageReader 3.4.2 (rev e1679305c8b8e2e6a38bf3fcb7699ba6a2b2860a)

#1  0x00005620f55a4341 in MainWindow::signalHandlerExec(int, bool) ()
#2  0x00005620f55a3df2 in MainWindow::signalHandler(int) ()
[...]

PR #708, opened 2025-03-07
added files to create AppImage

efa:
this should fix https://github.com/manisandro/gImageReader/issues/688#issuecomment-2698998502

Note: I used a work around in AppImage creation, manually including the two files listed at:
https://github.com/manisandro/gImageReader/issues/707#issuecomment-2704548847
this increase a lot the creation time, sorry.

The generated AppImage should work on any distro newer than the one the binary and AppImage is created on (or with glibc newer that the build host). As now I tryed:
gImageReader_3.4.2_2025-03-07_Linux_x86_64_64bit.AppImage  47 MB
on Debian12 only, but it is an AppImage rule

### DIFF
--- a/qt/src/MainWindow.cc
+++ b/qt/src/MainWindow.cc
@@ -621,7 +621,7 @@ void MainWindow::dictionaryAutoinstall() {
 					failed.append(filename);
 				}
 				popState();
-				dictPos += dictMatch.capturedEnd();
+				dictPos = dictMatch.capturedEnd();
 			}
 		}
 


manisandro:
Thanks! Would it be possible to also add a workflow to build the appimage automatically as a CI job?

efa:
are speaking about "GitHub Actions: build and test multi-platform", or another external service?

manisandro:
Yes Github Actions - a workflow similar to [1].

[1] https://github.com/manisandro/gImageReader/blob/master/.github/workflows/build.yml

efa:
I have no experience with Github actions, but it's something that I have in my todo list also for my applications.
I'd say it's the right time to study how it works and try it

efa:
If you give me some hints with Github actions, I will try

PR #713, opened 2025-04-15
chore: Improve code robustness and maintainability.

GongHeng2017:
1. Some variable can be declared as pointer to const.
2. The scope of the variable 'targetLine' can be reduced.
3. Local variable 'bottomLeft' shadows outer variable


GongHeng2017:
@manisandro

PR #728, opened 2025-08-18
Create Appimage

efa:
I dig a little in Github action to generate a CI binaries. I take inspiration from your yml file, thanks. Here the results:
https://github.com/efa/gImageReader/releases/tag/ci-latest

Note1:
As Github has image for Ubuntu and not for Debian, and I cannot find some QT6*dev packages (maybe freespell), I had to switch to this list of deps with Qt5:
https://github.com/manisandro/gImageReader/blob/master/packaging/debian/control
and not this one with Qt6:
https://salsa.debian.org/debian/gimagereader/-/blob/main/debian/control

Note 2:
I do not know why, but as now:
https://github.com/manisandro/gImageReader/blob/master/.github/workflows/build.yml
always fail for some Fedora deps error:
https://github.com/efa/gImageReader/actions/runs/17052058157/job/48341834644

This is not related to my new file:
https://github.com/efa/gImageReader/blob/appimage/.github/workflows/linux.yml

Note 3:
The similar old PR: https://github.com/manisandro/gImageReader/pull/708
is now overcome and closed


manisandro:
Thanks a lot!

The Fedora dependency issues are being resolved.

Regarding the appimage workflow, would it be possible to also build the qt variant instead of only gtk? Which qt6 dependencies are missing from debian/ubuntu exactly?

efa:
I made the changes to support generation of both GTK and QT5 AppImage (mainly name the package with the suffix -gtk or -qt5).

The script is ready for QT6 too, and work locally on my Debian13.
But github action Ubuntu latest is 24.04 that apparently miss libqtspell-qt6-dev to wont build the QT6 target.

Anyway I left commented in:
https://github.com/efa/gImageReader/blob/appimage/.github/workflows/linux.yml
the QT6 part if you find a solution

manisandro:
Sorry this fell of my radar. If you have no objections I would try to use a fedora base image to get the qt6 variant building, since qt5 is pretty obsolete nowadays.

efa:
apparently github support only Ubuntu as Linux distro, see:
https://github.com/actions/runner-images?tab=readme-ov-file#available-images

efa:
Unfortunately, recent distros no longer support QT5 (missing package "libqtspell-qt5-dev"), so you have to use two different distros to generate QT5 and QT6.

I looked up how to use Docker to generate AppImages on a distro newer than Ubuntu 24.04. I tested it with Debian 13, which I know well because FUSE must be enabled.
I created an additional file:
https://github.com/efa/gImageReader/blob/appimage/.github/workflows/docker-debian13.yml
which generates the GTK and QT6 packages, while the old one remain for QT5.
I just need to figure out how to make the kernel version path `uname -r` a parameter in Docker run.
Here the results:
https://github.com/efa/gImageReader/releases/tag/ci-latest

manisandro:
See [1], the outer container is indeed ubuntu, but you can run a fedora container within that container.

[1] https://github.com/manisandro/gImageReader/blob/master/.github/workflows/build.yml#L7

efa:
in the new file:
https://github.com/efa/gImageReader/blob/appimage/.github/workflows/docker-debian13.yml
I made use of docker

shuvashish76:
If anyone interested **unofficial** AppImage is available at https://github.com/pkgforge-dev/gImageReader-AppImage by Package Forge devs.

